PR #6, opened 2018-12-27
Db implementation

Nashatyrev:
This is ongoing PR for db


db/core/src/main/java/org/ethereum/beacon/db/source/HoleyList.java
```diff
@@ -0,0 +1,59 @@
...
+   * Returns element at index <code>idx</code>
+   * Empty instance is returned if no element with this index
+   */
+  Optional<V> get(long idx);
+
+  /**
+   * Puts element with index <code>size()</code>
+   */
+  default void add(V value) {
+    put(size(), value);
```
zilm13: I'd return result here

build.gradle
```diff
@@ -14,6 +14,8 @@ allprojects {
   targetCompatibility = '1.8'
 
   dependencies {
+    implementation 'com.google.code.findbugs:jsr305'
```
zilm13: What is it for?

chain/src/main/java/org/ethereum/beacon/chain/storage/AbstractChainStorage.java
```diff
@@ -0,0 +1,47 @@
+package org.ethereum.beacon.chain.storage;
+
+import org.ethereum.beacon.chain.storage.BeaconBlockStorage;
+import org.ethereum.beacon.chain.storage.BeaconStateStorage;
+import org.ethereum.beacon.chain.storage.BeaconTupleStorage;
+import org.ethereum.beacon.chain.storage.ChainStorage;
+import org.ethereum.beacon.chain.storage.impl.BeaconTupleStorageImpl;
+import org.ethereum.beacon.db.Database;
+
+public abstract class AbstractChainStorage implements ChainStorage {
```
zilm13: Doesn't look like something abstract.
Though I'm ok to keep naming simple if it's in specific package.

chain/src/main/java/org/ethereum/beacon/chain/storage/ChainStorage.java
```diff
@@ -0,0 +1,17 @@
+package org.ethereum.beacon.chain.storage;
+
+import org.ethereum.beacon.db.Database;
+
+/**
+ * Created by Anton Nashatyrev on 27.12.2018.
+ */
+public interface ChainStorage {
```
zilm13: Doesn't look like something abstract.
Though I'm ok to keep naming simple if it's in specific package.

db/core/src/main/java/org/ethereum/beacon/db/source/HoleyList.java
Nashatyrev: What result?

build.gradle
Nashatyrev: `@Nonnull`, `@Nullable`

chain/src/main/java/org/ethereum/beacon/chain/storage/ChainStorage.java
Nashatyrev: Not sure what you mean here... Let's talk on tomorrow call

build.gradle
zilm13: Aaah, ok

db/core/src/main/java/org/ethereum/beacon/db/source/HoleyList.java
zilm13: Could it update old one, or it's always new?

chain/src/main/java/org/ethereum/beacon/chain/storage/ChainStorage.java
zilm13: I mean `ChainStorage` sounds abstract but `BeaconBlockStorage` etc are very specific.  So smth like `BeaconChainStorage` and `AbstractBeaconChainStorage` could be better.
But anyway I'm ok to keep it shorter like it is now.

db/core/src/main/java/org/ethereum/beacon/db/source/HoleyList.java
Nashatyrev: Yep, it's always new

Issue #9, opened 2018-12-29
Check proof of possession and withdrawal credentials when processing deposits


mkalinin:
### What's wrong
Proof of possession and withdrawal credentials are not checked when deposit is being processed.

### Proposal
Proof of possession check should be a part of verification logic, blocks that includes deposits with wrong proofs of possession should be considered as invalid. Initial deposits with wrong proof should be discarded silently with INFO message in the log. 

Withdrawal credentials does make sense only for top ups. Hence, it should be a part of verification logic with same behavior as for proofs of possession.

Nashatyrev:
Block deposits validation is now implemented within `SpecHelpers.process_deposit`
Separate issue regarding initial deposit validation

mkalinin:
Addressed by https://github.com/harmony-dev/beacon-chain-java/pull/98

PR #10, opened 2018-12-29
Initial state transition

mkalinin:
### What's done
- Addresses [On startup](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#on-startup) section in the spec
- ~~Committees initialization is not yet implemented~~ and won't be cause it's gonna be removed soon, see https://github.com/ethereum/eth2.0-specs/pull/408
- Proofs of possession for initial deposits are not checked, see https://github.com/harmony-dev/beacon-chain-java/issues/9
- Tests are pending https://github.com/harmony-dev/beacon-chain-java/issues/11, https://github.com/harmony-dev/beacon-chain-java/issues/12


core/src/main/java/org/ethereum/beacon/core/BeaconState.java
```diff
@@ -19,21 +25,40 @@
  *     in the spec</a>
  */
 public class BeaconState implements Hashable<Hash32> {
+
+  /** Max number of RANDAO mixes kept by {@link #latestRandaoMixes}. */
+  public static final int LATEST_RANDAO_MIXES_LENGTH = 1 << 13; // 8192
```
Nashatyrev: These constants also look like a part of `ChainSpec`

PR #13, opened 2019-01-10
Chain spec

mkalinin:
### What's done
`BeaconChainSpec` class with static variables replaced with `ChainSpec` interface with its default implementation that uses constant values from the spec.

### Implementation details
`ChainSpec` interface extends several parameter interfaces like `MiscParameters`. Each parameter interface reflects own [constant section](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#constants) in the spec.

`ChainSpec` does not include:
- validator status flags defined in `ValidatorStatusFlags`
- validator registry delta flags defined in `ValidatorRegistryDeltaFlags`
- signature domains defined in `SignatureDomains`

These constants are not meant to be changed across different chain instances.


consensus/src/main/java/org/ethereum/beacon/consensus/transition/InitialStateTransition.java
```diff
@@ -67,9 +69,11 @@ public BeaconState apply(BeaconBlock block, BeaconState state) {
 
     // Randomness and committees
     builder
-        .withLatestRandaoMixes(nCopies(BeaconState.LATEST_RANDAO_MIXES_LENGTH, Hash32.ZERO))
+        .withLatestRandaoMixes(
+            nCopies(chainSpec.getLatestRandaoMixesLength().getIntValue(), Hash32.ZERO))
         .withLatestVdfOutputs(
-            nCopies(BeaconState.LATEST_RANDAO_MIXES_LENGTH / Epoch.LENGTH, Hash32.ZERO))
+            nCopies(
+                chainSpec.getLatestRandaoMixesLength().getIntValue() / Epoch.LENGTH, Hash32.ZERO))
```
Nashatyrev: I suppose `Epoch.LENGTH` should also be in the `ChainSpec` ?

types/src/main/java/tech/pegasys/artemis/util/uint/UInt64.java
```diff
@@ -38,6 +39,10 @@ public long getValue() {
     return value;
   }
 
+  public int getIntValue() {
+    return (int) value;
+  }
```
Nashatyrev: Ugh, scary thing :)
I would add the assertion `value >= 0 && value <= Integer.MAX_VALUE`

consensus/src/main/java/org/ethereum/beacon/consensus/transition/InitialStateTransition.java
mkalinin: Yep, it's just in the next section of constants. Not yet covered

PR #15, opened 2019-01-14
Fetaure/state interface

Nashatyrev:
`BeaconState` as interfaces


consensus/src/main/java/org/ethereum/beacon/consensus/transition/InitialStateTransition.java
```diff
@@ -1,32 +1,35 @@
...
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.nCopies;
+
 /**
  * Produces initial beacon state.
  *
  * <p>Requires input {@code block} to be a Genesis block, {@code state} parameter is ignored.
- * Preferred input for {@code state} parameter is {@link BeaconState#EMPTY}.
+ * Preferred input for {@code state} parameter is EMPTY.
```
mkalinin: Better to refer to `BeaconState#getEmpty()` here.

consensus/src/main/java/org/ethereum/beacon/consensus/transition/InitialStateTransition.java
```diff
@@ -51,10 +54,10 @@ public BeaconState apply(BeaconBlock block, BeaconState state) {
 
     ChainStart chainStart = depositContract.getChainStart();
 
-    Builder builder = BeaconState.Builder.createEmpty();
+    MutableBeaconState initialState = MutableBeaconState.createNew();
```
mkalinin: What do you think about keeping `BeaconStateImpl.Builder` which is either extends or delegating to `MutableBeaconState`?

core/src/main/java/org/ethereum/beacon/core/state/BeaconStateImpl.java
```diff
@@ -0,0 +1,428 @@
...
+    this.latestAttestations = latestAttestations;
+    this.batchedBlockRoots = batchedBlockRoots;
+    this.latestDepositRoot = latestDepositRoot;
+    this.depositRootVotes = depositRootVotes;
+
+    validate();
+  }
+
+  @Override
+  public BeaconState validate() {
```
mkalinin: IMO, for `MutableBeaconState` it's better to run validation in each setter separately. Client might just forget to call `validate()` after all the changes.

core/src/main/java/org/ethereum/beacon/core/MutableBeaconState.java
```diff
@@ -0,0 +1,185 @@
...
+import org.ethereum.beacon.core.state.*;
+import tech.pegasys.artemis.ethereum.core.Hash32;
+import tech.pegasys.artemis.util.uint.UInt64;
+
+import java.util.List;
+
+public interface MutableBeaconState extends BeaconState {
+
+  static MutableBeaconState createNew() {
+    return new BeaconStateImpl();
```
mkalinin: `BeaconStateImpl` is package private. Not possible to call it here.

core/src/main/java/org/ethereum/beacon/core/state/BeaconStateImpl.java
Nashatyrev: Yep, I don't like this either too much. 
I'm not sure the State can be consistent on each `setXXX` call

consensus/src/main/java/org/ethereum/beacon/consensus/transition/InitialStateTransition.java
Nashatyrev: If extending what should its purpose? 
If delegating then it's another bunch of boilerplate code :/
Also I would like to hide the `BeaconStateImpl` class (of course the Builder can be independent of it)

mkalinin: >  If delegating then it's another bunch of boilerplate code :/

Yep, there could be toughs in making changes to `MutableBeaconState` and _not forgetting_ about a `Builder`. Builders are good that they either produce a valid object or throwing an error. But we may keep going without a builder for now.

core/src/main/java/org/ethereum/beacon/core/state/BeaconStateImpl.java
mkalinin: For sure, it can't. But you may check only an argument that is changed by particular setter. I.e. check `XXX` in case of `setXXX`.

PR #18, opened 2019-01-14
SSZ hash

zilm13:
Not yet ready for merge. Not working. But will look similar to this

### DIFF
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
@@ -1,4 +1,4 @@
-package org.ethereum.beacon.types.ssz;
+package org.ethereum.beacon.ssz;
 
 import net.consensys.cava.bytes.Bytes;
 import org.ethereum.beacon.ssz.ConstructorObjCreator;
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
@@ -1,77 +1,14 @@
 package org.ethereum.beacon.ssz;
 
-import net.consensys.cava.bytes.Bytes;
-import org.ethereum.beacon.ssz.ConstructorObjCreator;
-import org.ethereum.beacon.ssz.SSZAnnotationSchemeBuilder;
-import org.ethereum.beacon.ssz.SSZCodecHasher;
-import org.ethereum.beacon.ssz.SSZCodecResolver;
-import org.ethereum.beacon.ssz.SSZHashSerializer;
-import org.ethereum.beacon.ssz.SSZModelCreator;
-import org.ethereum.beacon.ssz.SSZModelFactory;
-import org.ethereum.beacon.ssz.SSZSerializer;
-import org.ethereum.beacon.ssz.SSZSerializerBuilder;
-import org.ethereum.beacon.ssz.SettersObjCreator;
-import tech.pegasys.artemis.util.bytes.BytesValue;
-import java.util.function.Function;
-
 /**
- * <p>SSZ Hashing helper made according to the following specs:
- * <a href="https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash">SSZ Tree Hash</a></p>
- *
- * <p>It's similar to {@link org.ethereum.beacon.ssz.SSZHasher}, but made to work with BytesValue</p>
+ * Hasher
+ * @param <R> return type
  */
-public class Hasher {
-  private SSZSerializer hasher;
-
-  private Function<BytesValue, BytesValue> hashFunction;
-
+public interface Hasher<R> {
   /**
-   * Creates instance of SSZ Hasher with
-   * @param builder        Almost baked builder, without {@link SSZSerializerBuilder#build()} method being called
-   * @param hashFunction   Function that will be used for hashing
+   * Calculatesh hash of object
+   * @param input   Input object to calculate hash of
+   * @return  Hash
    */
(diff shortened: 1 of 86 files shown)


zilm13:
@mkalinin ready for review

ssz/src/main/java/org/ethereum/beacon/ssz/SSZSerializerBuilder.java
```diff
@@ -50,6 +50,16 @@ public SSZSerializerBuilder(SSZSchemeBuilder schemeBuilder, SSZCodecResolver cod
     this.sszSerializer = new SSZSerializer(schemeBuilder, codecResolver, sszModelFactory);
   }
 
+  public SSZSerializerBuilder initWith(SSZSerializer sszSerializer, SSZCodecResolver codecResolver) {
```
mkalinin: Maybe keep builder simple and conventional? I mean use plain `withSszSerializer`, `withCodecResolver`, etc. methods?

ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
```diff
@@ -0,0 +1,77 @@
...
+import tech.pegasys.artemis.util.bytes.BytesValue;
+import java.util.function.Function;
+
+/**
+ * <p>SSZ Hashing helper made according to the following specs:
+ * <a href="https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash">SSZ Tree Hash</a></p>
+ *
+ * <p>It's similar to {@link org.ethereum.beacon.ssz.SSZHasher}, but made to work with BytesValue</p>
+ */
+public class Hasher {
```
mkalinin: Maybe stay with `SSZHasher` only then? Don't think we need both.

ssz/src/main/java/org/ethereum/beacon/ssz/SSZCodecHasher.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * <p>Implementation of {@link SSZCodecResolver} which implements SSZ Hash function</p>
+ * <p>For more info check <a href="https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash">SSZ Tree hash</a></p>
+ */
+public class SSZCodecHasher implements SSZCodecResolver {
```
mkalinin: I don't think that this is a right place for hashing stuff. I mean that resolver doesn't meant to be a good place for extensions like this one. It would be great to have a codec pipeline in `SSZSerializer` that could be extended with any class that for example implements `SSZCodec`. Even `SSZCodecResolver` could be aggregated by some `PrimitiveSSZCodec implements SSZCodec`. Then to instantiating a `Hasher` would only require one more call to `SSZSerializerBuilder` comparing with regular `SSZSerializer` instantiation.

But this is just an idea. Need to check how it's applicable to the implementation we have.

ssz/src/main/java/org/ethereum/beacon/ssz/SSZHasher.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.function.Function;
+
+/**
+ * <p>SSZ Hashing helper made according to the following specs:
+ * <a href="https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash">SSZ Tree Hash</a></p>
+ *
+ * <p>It's based on {@link SSZSerializer}, so for setup help, check its Javadoc
+ * and {@link SSZSerializerBuilder} documentation</p>
+ */
+public class SSZHasher {
```
mkalinin: What do you think about adding a parameter that specifies a type of the output hash value?

types/src/main/java/org/ethereum/beacon/types/ssz/SSZBytesValue.java
```diff
@@ -179,15 +179,15 @@ public Object decode(SSZSchemeBuilder.SSZScheme.SSZField field, BytesSSZReaderPr
   }
 
   static class BytesType {
-    Integer size = null;
+    final Integer size;
+    public static BytesType FLEXIBLE = new BytesType(null);
```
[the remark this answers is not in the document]
mkalinin: wouldn't `DYNAMIC` be a better name?

ssz/src/main/java/org/ethereum/beacon/ssz/SSZHasher.java
zilm13: done

ssz/src/main/java/org/ethereum/beacon/ssz/Hasher.java
zilm13: done

ssz/src/main/java/org/ethereum/beacon/ssz/SSZSerializerBuilder.java
zilm13: rebuilt

ssz/src/main/java/org/ethereum/beacon/ssz/SSZCodecHasher.java
zilm13: > It would be great to have a codec pipeline in SSZSerializer that could be extended with any class that for example implements SSZCodec

Currently, for example, calculating hash for container and list differs like Africa and Europe. You are trying to find object-oriented way which will combine just 2 tasks, and any of them could change dramatically tomorrow. Current solution is a simple way to deliver it fast, and any changes to hash specs could make both mine and your solutions impossible to work without hacks. I'd stay with simple and fast thing here.

PR #20, opened 2019-01-15
Block verification

mkalinin:
### What's done
Adds pre and post processing verification:
- state matching when block is processed
- proposer signature checks
- RANDAO checks
- proposer slashing checks
- casper slashing checks
- attestation checks
- ~~[WIP] deposit checks~~ tightly coupled with Deposit contract logic, moved to an issue https://github.com/harmony-dev/beacon-chain-java/issues/24
- exit checks


core/src/main/java/org/ethereum/beacon/core/spec/ChainSpec.java
```diff
@@ -43,7 +43,7 @@ public UInt64 getBeaconChainShardNumber() {
         }
 
         @Override
-        public UInt64 getMaxCasperVotes() {
+        public int getMaxCasperVotes() {
```
zilm13: Why? Should we?

mkalinin: This constant is applied to a list/array size which have `int` type in Java. This is what stays behind this change.

zilm13: We may need to use any other constant could some day in Java code.

Nashatyrev: I'm going to review all those casts and types at some point

mkalinin: This particular constant limits a size of `SlashableVoteData.custodyBit0Indices + SlashableVoteData.custodyBit1Indices`. And not gonna be used at some other place.

PR #25, opened 2019-01-16
Feture/next slot

Nashatyrev:
Add `BeaconStateEx` wrapper to handle latest block for `state.latest_block_roots` modification. 
Since the block root for a slot state is added only on the next slot transition we need a way to keep the added block till the next transition. 
The general idea was to add this field to the `BeaconState` (without including to the hash calculation), it the intention is to keep the `BeaconState` canonical.

### DIFF
--- a/consensus/src/main/java/org/ethereum/beacon/consensus/StateTransition.java
+++ b/consensus/src/main/java/org/ethereum/beacon/consensus/StateTransition.java
@@ -5,5 +5,5 @@ import org.ethereum.beacon.core.BeaconState;
 
 public interface StateTransition<S> {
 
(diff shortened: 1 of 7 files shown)


consensus/src/main/java/org/ethereum/beacon/consensus/StateTransition.java
```diff
@@ -5,5 +5,5 @@
 
 public interface StateTransition<S> {
 
-  BeaconState apply(BeaconBlock block, S state);
+  S apply(BeaconBlock block, S state);
```
mkalinin: What a reasonable change!

PR #29, opened 2019-01-17
Feature/epoch trans

Nashatyrev:
Initial translation of Epoch transition from Python to Java
Not tested in any ways for now


core/src/main/java/org/ethereum/beacon/core/state/ValidatorRecord.java
```diff
@@ -258,6 +260,11 @@ public Builder withStatusFlags(UInt64 statusFlags) {
       return this;
     }
 
+    public Builder withStatusFlags(Function<UInt64, UInt64> statusFlagsUpdater) {
```
mkalinin: Why not just update flags with pre-calculated value?

types/src/main/java/org/ethereum/beacon/types/Ether.java
```diff
@@ -27,6 +27,14 @@ public static Ether valueOf(long unsignedValue) {
     return new Ether(UInt64.valueOf(unsignedValue));
   }
 
+  public static Ether ofGWeis(UInt64 gWeis) {
```
mkalinin: I think we need too add a separate type for `GWei` later on.

consensus/src/main/java/org/ethereum/beacon/consensus/transition/BlockTransition.java
```diff
@@ -0,0 +1,167 @@
...
+       For each casper_slashing in block.body.casper_slashings:
+       Let slashable_vote_data_1 = casper_slashing.slashable_vote_data_1.
+       Let slashable_vote_data_2 = casper_slashing.slashable_vote_data_2.
+       Let indices(slashable_vote_data) =
+          slashable_vote_data.custody_bit_0_indices + slashable_vote_data.custody_bit_1_indices.
+       Let intersection = [x for x in indices(slashable_vote_data_1) if x in indices(slashable_vote_data_2)].
+       For each validator index i in intersection run penalize_validator(state, i)
+           if state.validator_registry[i].penalized_slot > state.slot.
+    */
+    for (CasperSlashing casper_slashing : block.getBody().getCasperSlashings()) {
```
mkalinin: There are `SpecHelpers.indices` and `SpecHelpers.intersection` methods to make this shorter

consensus/src/main/java/org/ethereum/beacon/consensus/transition/BlockTransition.java
```diff
@@ -0,0 +1,167 @@
...
+           proof_of_possession=deposit.deposit_data.deposit_input.proof_of_possession,
+           withdrawal_credentials=deposit.deposit_data.deposit_input.withdrawal_credentials,
+           randao_commitment=deposit.deposit_data.deposit_input.randao_commitment,
+           custody_commitment=deposit.deposit_data.deposit_input.custody_commitment,
+       )
+    */
+    final ValidatorRegistryUpdater registryUpdater =
+        ValidatorRegistryUpdater.fromState(state, spec);
+    for (Deposit deposit : block.getBody().getDeposits()) {
+      registryUpdater.processDeposit(deposit);
```
mkalinin: Do we want to maintain both `SpecHelpers` and `RegistryUpdater`? Maybe abandon `ValidatorRegistryUpdater` for a while and handle everything in `SpecHelpers`?

core/src/main/java/org/ethereum/beacon/core/state/ValidatorRecord.java
Nashatyrev: Because we need to *modify* existing flag:
```python
     validator.status_flags |= INITIATED_EXIT
```

types/src/main/java/org/ethereum/beacon/types/Ether.java
Nashatyrev: Right now doing it

consensus/src/main/java/org/ethereum/beacon/consensus/transition/BlockTransition.java
Nashatyrev: Yep, makes sense for now

core/src/main/java/org/ethereum/beacon/core/state/ValidatorRecord.java
mkalinin:
```java
validtor.withStatusFlags(validator.getStatusFlags().or(INITIATED_EXIT));
```
looks pretty similar to
```java
validtor.withStatusFlags(flags -> flags.or(INITIATED_EXIT));
```

Anyway, not a big deal

types/src/main/java/org/ethereum/beacon/types/Ether.java
mkalinin: 👍

core/src/main/java/org/ethereum/beacon/core/state/ValidatorRecord.java
Nashatyrev: This is `Builder` instance passed to this method 
```
withValidatorRecord(ValidatorIndex idx,
      Consumer<ValidatorRecord.Builder> validatorUpdater)
```
it has no `getStatusFlags()`
Anyway, not a big deal :)

mkalinin: Oh, got it!

consensus/src/main/java/org/ethereum/beacon/consensus/transition/BlockTransition.java
Nashatyrev: Done

PR #30, opened 2019-01-17
specHelpers: Fixed verify_slashable_vote_data inputs according to spe…

zilm13:
See https://github.com/ethereum/eth2.0-specs/pull/465


mkalinin:
To complete this fix `state.slot` that is passed to `get_domain` function in `AttestationVerifier`, `PropserSlashingVerifier`, `ExitVerifier` should also be replaced with slot taken from verifying structure.

PR #36, opened 2019-01-21
Validator service

mkalinin:
### What's done
Validator service design with related interfaces and dependencies.


consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -12,11 +12,14 @@
 import java.util.List;
```
Nashatyrev: May be let's finally decide on style enforcement and if decided to use strict format then reformat all the classes separately?

pending/src/main/java/org/ethereum/beacon/pending/ObservableChainState.java
```diff
@@ -0,0 +1,38 @@
...
+import org.ethereum.beacon.core.BeaconBlock;
+import org.ethereum.beacon.core.BeaconState;
+
+/**
+ * An observable chain state includes:
+ *
+ * <ul>
+ *   <li>{@link BeaconBlock} that is a head of the chain;
+ *   <li>{@link BeaconState} after recent slot has been processed;
+ *   <li>{@link PendingState} constantly updating with latest messages from the net.
```
Nashatyrev: Does this mean `PendingState` instance is mutable? If not, may be rephrase?

chain/src/main/java/org/ethereum/beacon/chain/BeaconChainListener.java
```diff
@@ -0,0 +1,8 @@
+package org.ethereum.beacon.chain;
+
+import java.util.concurrent.CompletableFuture;
+
+/** A stub for beacon chain listener. */
+public interface BeaconChainListener {
```
Nashatyrev: Can be replaced by rx `Publisher` later (streams have a notion of `Mono` - 0|1 element stream)

pow/src/main/java/org/ethereum/beacon/pow/PoWChainListener.java
```diff
@@ -0,0 +1,14 @@
+package org.ethereum.beacon.pow;
+
+import java.util.concurrent.CompletableFuture;
+
+/** A dummy interface that could replaced with any more convenient pattern later. */
+public interface PoWChainListener {
```
Nashatyrev: Can also be replaced by `Publisher` later

validator/src/main/java/org/ethereum/beacon/validator/SingleValidatorService.java
```diff
@@ -0,0 +1,187 @@
...
+
+  /** FIXME: a stub for subscription. */
+  private void subscribeToDepositRootUpdates(Consumer<Hash32> payload) {}
+
+  /** FIXME: a stub for subscription. */
+  private void waitForTheFirstHead(Consumer<ObservableChainState> payload) {}
+
+  private void waitForBeaconChainSync() {
+    try {
+      beaconChainListener.getSyncDone().get();
```
Nashatyrev: (I suspect this is stub for now) Blocking operations should be avoided

PR #37, opened 2019-01-22
Reactive streams integration to BeaconChain

Nashatyrev:
Concept of using reactive streams in the library


chain/src/main/java/org/ethereum/beacon/chain/BeaconChain.java
```diff
@@ -1,6 +1,16 @@
 package org.ethereum.beacon.chain;
 
+import org.ethereum.beacon.chain.storage.BeaconTuple;
+import org.ethereum.beacon.core.BeaconState;
+import org.reactivestreams.Publisher;
+
 public interface BeaconChain {
 
+  Publisher<BeaconChainHead> getHeadStream();
```
mkalinin: What do you think about moving streams into a separate interface?

PR #38, opened 2019-01-22
Fork choice

zilm13:
Help me. Guide me.


chain/src/main/java/org/ethereum/beacon/chain/DefaultBeaconChain.java
```diff
@@ -120,6 +116,19 @@ public synchronized void insert(BeaconBlock block) {
     BeaconState parentState = pullParentState(block);
     BeaconState slotTransitedState = slotTransition.apply(block, parentState);
 
+    // update head
+    headFunction
```
mkalinin: Maybe writing this logic within several lines would make it more readable?

chain/src/main/java/org/ethereum/beacon/chain/GhostHeadFunction.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * The beacon chain fork choice rule is a hybrid that combines justification and finality with
+ * Latest Message Driven (LMD) Greediest Heaviest Observed SubTree (GHOST). For more info check <a
+ * href="https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beacon-chain-fork-choice-rule">Beacon
+ * chain fork choice rule</a>
+ */
+public class GhostHeadFunction implements HeadFunction {
```
mkalinin: I'd specify which of `GHOST` algorithms is used.

chain/src/main/java/org/ethereum/beacon/chain/GhostHeadFunction.java
```diff
@@ -0,0 +1,222 @@
...
+  //        return len([target for target in attestation_targets if
+  //                    get_ancestor(store, target, block.slot) == block])
+  //
+  //    head = start
+  //    while 1:
+  //        children = get_children(head)
+  //        if len(children) == 0:
+  //            return head
+  //        head = max(children, key=get_vote_count)
+  private BeaconBlock lmd_ghost(BeaconTuple startTuple) {
```
mkalinin: Maybe move this one and other functions defined below to `SpecHelpers`? As their source is the spec. `store` parameter could be replaced with `Function<Hash32, List<BeaconBLock>>`.

chain/src/main/java/org/ethereum/beacon/chain/GhostHeadFunction.java
```diff
@@ -0,0 +1,222 @@
...
+        blockStorage
+            .getSlotCanonicalBlock(lastJustifiedSlot.getValue())
+            .map(tupleStorage::get)
+            .orElseThrow(() -> new RuntimeException("No justified head found"));
+    BeaconBlock newHead =
+        justifiedTuple
+            .map(this::lmd_ghost)
+            .orElseThrow(() -> new RuntimeException("No justified head found"));
+    if (!newHead.getHash().equals(curHead)) {
+      blockStorage.reorgTo(newHead.getHash());
```
mkalinin: IMO, we should get rid of reorgs in block storage and calculate head on the fly after restart. This, also, implies other changes on block storage:
- get rid of canonical chain flat;
- add a state flag with three states: NONE, JUSTIFIED, FINALIZED.

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1194,6 +1197,114 @@ public void checkShardRange(UInt64 shard) {
...
+   *     use the one the validator v observed first.
+   */
+  private BeaconBlock get_latest_attestation_target(
+      ValidatorRecord validatorRecord,
+      Function<ValidatorRecord, Attestation> get_latest_attestation,
+      Function<Hash32, Optional<BeaconBlock>> getBlock) {
+    Attestation latest = get_latest_attestation.apply(validatorRecord);
+    return getBlock
+        .apply(latest.getData().getJustifiedBlockRoot())
+        .orElseThrow(
```
mkalinin: I don't think this is exceptional case. There could be a recent network attestation which refers to not yet imported block, rare case for justified block, though. I think in this case an attestation for prior slot should be picked.

chain/src/main/java/org/ethereum/beacon/chain/storage/impl/BeaconBlockStorageImpl.java
```diff
@@ -200,16 +164,78 @@ public void remove(@Nonnull Hash32 key) {
...
+          blockIndex
+              .get(i)
+              .flatMap(slotBlocks -> Optional.of(slotBlocks.setJustifiedHash(justifiedHash)));
+      if (updated.isPresent()) {
+        blockIndex.put(i, updated.get());
+      }
+    }
+  }
+
+  public void addFinalizedHash(Hash32 finalizedHash) {
```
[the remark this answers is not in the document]
zilm13: this will broke maxSlot

chain/src/main/java/org/ethereum/beacon/chain/storage/impl/BeaconBlockStorageImpl.java
```diff
@@ -200,16 +164,78 @@ public void remove(@Nonnull Hash32 key) {
...
-      int newIdx = slotBlocks.getCanonicalIndex() - (slotBlocks.getCanonicalIndex() < idx ? 0 : 1);
       ArrayList<Hash32> newBlocks = new ArrayList<>(slotBlocks.getBlockHashes());
       newBlocks.remove(idx);
-      blockIndex.put(block.get().getSlot().getValue(), new SlotBlocks(newBlocks, newIdx));
+      blockIndex.put(block.get().getSlot().getValue(), new SlotBlocks(newBlocks, justifiedHash, finalizedHash));
+    }
+  }
+
+  @Override
+  public List<BeaconBlock> getChildren(@Nonnull Hash32 parent, int limit) {
```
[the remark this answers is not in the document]
zilm13: So, what's wrong with current implementation?

chain/src/main/java/org/ethereum/beacon/chain/LMDGhostHeadFunction.java
```diff
@@ -0,0 +1,154 @@
...
+            .map(BeaconTuple::getState)
+            .orElseThrow(() -> new RuntimeException("No justified tuple found"));
+    Flux.from(beaconChain.getBlockStatesStream())
+        .doOnNext(beaconTuple -> state = beaconTuple.getState())
+        .subscribe();
+  }
+
+  @Override
+  public BeaconBlock getHead() {
+    Optional<Hash32> justifiedHash = blockStorage.getSlotJustifiedBlock(blockStorage.getMaxSlot());
```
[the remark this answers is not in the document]
zilm13: this state has not get enough votes yet, so it is not justified yet if I get it correctly.

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1194,6 +1197,114 @@ public void checkShardRange(UInt64 shard) {
...
+
+    return res;
+  }
+
+  /**
+   * Let get_ancestor(store, block, slot) be the ancestor of block with slot number slot. The
+   * get_ancestor function can be defined recursively as def get_ancestor(store, block, slot):
+   * return block if block.slot == slot else get_ancestor(store, store.get_parent(block), slot).
+   */
+  private BeaconBlock get_ancestor(
```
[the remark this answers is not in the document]
zilm13: why? targets in incoming data could be anywhere deeper
and it's exactly the code from spec

zilm13:
@mkalinin ready for review

chain/src/main/java/org/ethereum/beacon/chain/storage/impl/BeaconBlockStorageImpl.java
```diff
@@ -182,9 +139,14 @@ public void put(@Nonnull Hash32 key, @Nonnull BeaconBlock newBlock) {
     }
 
     rawBlocks.put(key, newBlock);
-    blockIndex.update(newBlock.getSlot().getValue(),
-        blocks -> blocks.addBlock(newBlock.getHash(), isCanonical),
-        () -> new SlotBlocks(newBlock.getHash(), isCanonical));
+    SlotBlocks slotBlocks =
```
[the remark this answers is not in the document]
zilm13: we are not storing in current implementation
only in slots which contains justified we are storing which one is justified

chain/src/main/java/org/ethereum/beacon/chain/storage/impl/BeaconBlockStorageImpl.java
```diff
@@ -2,142 +2,109 @@
...
-    private static final int NO_CANONICAL = -1;
 
     private final List<Hash32> blockHashes;
-    // -1: no canonical block
-    private final int canonicalIndex;
+    private final Hash32 justifiedHash;
+    private final Hash32 finalizedHash;
+
+    SlotBlocks(Hash32 blockHash) {
+      this(singletonList(blockHash), null, null);
```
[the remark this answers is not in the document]
zilm13: if we next are serializing it directly, it consumes all 32 bytes.
plus why to use it when we have null?
it's ok to agree on special value when you don't have it, say -1 for UInt. But when you have it already, why to add something to agreement?

chain/src/main/java/org/ethereum/beacon/chain/storage/impl/BeaconBlockStorageImpl.java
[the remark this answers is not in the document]
mkalinin: Ah, I see.
`new SlotBlocks(newBlock.getHash(), newBlock.getHash(), newBlock.getHash())`
what is this for?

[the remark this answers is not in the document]
zilm13: genesis block

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -51,9 +55,24 @@
...
   public SpecHelpers(ChainSpec spec) {
     this.spec = spec;
+    this.objectHasher = input -> Hash32.ZERO;
+  }
+
+  /** Builds objectHasher with {@link #hash(BytesValue)} as data hash function in objectHasher */
+  public SpecHelpers(
+      ChainSpec spec, Function<Function<BytesValue, Hash32>, Hasher<Hash32>> objectHasherBuilder) {
+    this.spec = spec;
+    this.objectHasher = objectHasherBuilder.apply(this::hash);
```
mkalinin: Could we explicitly call `SSZHasher.createXXX(Function<BytesValue, Hash32> hashFunction)` instead of this? What could the other application for this abstraction be?

zilm13: how could we pass this::hash which is non static this way?

mkalinin: It could be passed as an argument to a factory method, I think.

mkalinin: Something like this:
```this.objectHasher = SSZHasher.createXXX(this::hash);```

PR #40, opened 2019-01-23
Beacon block proposer

mkalinin:
### What's done
First iteration of block proposer logic. 

Also, includes:
- an update to `randao_reveal` logic https://github.com/ethereum/eth2.0-specs/pull/483
- update `eth1_data` data structures and logic


types/src/main/java/tech/pegasys/artemis/util/uint/UInt64.java
```diff
@@ -320,6 +321,13 @@ public Bytes8 toBytesBigEndian() {
     return Bytes8.wrap(array);
   }
 
+  public Bytes32 toBytes32BigEndian() {
```
Nashatyrev: I would better use `Bytes32.leftPad(UInt64.toBytesBigEndian())` since it explicitly specify how these 8 bytes are padded

mkalinin: Thanks 👍

Nashatyrev: I mean using this (`Bytes32.leftPad(UInt64.toBytesBigEndian())`) in code explicitly , since 
- this is very domain specific conversion for including this method to pretty abstract `UInt64`
- the name `toBytes32BigEndian` doesn't reflect semantics (the same ways as `toBytes` name)

mkalinin: Yep, agree

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
```diff
@@ -0,0 +1,177 @@
...
+public class BeaconChainProposerImpl implements BeaconChainProposer {
+
+  private SpecHelpers specHelpers;
+  private ChainSpec chainSpec;
+  private StateTransition<BeaconState> blockTransition;
+  private DepositContract depositContract;
+
+  public BeaconChainProposerImpl(
+      SpecHelpers specHelpers,
+      ChainSpec chainSpec,
```
Nashatyrev: BTW I've added `ChainSpec` as a member to `SpecHelpers`. 
I think it make sense to unite them into a single essence and not passing them both to every instance

Nashatyrev: Or even unite 3 things: `ChainSpec`, `SpecHelpers` and transitions, since they are all represent the Spec2.0

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
```diff
@@ -0,0 +1,177 @@
...
+    Eth1DataVote bestVote =
+        eth1DataVotes.stream()
+            .filter(eth1DataVote -> eth1DataVote.getVoteCount().equals(maxVotes))
+            .reduce((first, second) -> second)
+            .get();
+
+    // verify best vote data and return if verification passed,
+    // otherwise, return data from contract
+    return depositContract
+        .getDepositRoot(bestVote.getEth1Data().getBlockHash())
```
Nashatyrev: why not just `bestVote.getEth1Data().getDepositRoot()` ?

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
mkalinin: I don't think there is a big benefit in mixing transitions in. But definitely makes sense for `ChainSpec` and `SpecHelper`.

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
mkalinin: I think that proposer should get a proof that root posted by a 3rd party exists.

Nashatyrev: Oh, got it!

pow/src/main/java/org/ethereum/beacon/pow/DepositContract.java
```diff
@@ -15,9 +15,9 @@
 
   List<Deposit> peekDeposits(int count, Eth1Data eth1Data, UInt64 fromIndex);
 
-  Eth1Data getEth1Data(long followDistance);
+  Eth1Data getEth1DataByDistance(long distanceFromHead);
 
-  Optional<Hash32> getDepositRoot(Hash32 blockHash);
+  Optional<Eth1Data> getEth1DataByBlockHash(Hash32 blockHash);
```
Nashatyrev: I've added `hasDepositRoot(Hash32 blockHash, Hash32, depositRoot)` method. Since many deposit roots may exist in a bock

mkalinin: Yep, sounds reasonable.

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
```diff
@@ -0,0 +1,170 @@
...
+    List<ProposerSlashing> proposerSlashings =
+        operations.peekProposerSlashings(chainSpec.getMaxProposerSlashings());
+    List<CasperSlashing> casperSlashings =
+        operations.peekCasperSlashings(chainSpec.getMaxCasperSlashings());
+    List<Attestation> attestations =
+        operations.peekAggregatedAttestations(chainSpec.getMaxAttestations());
+    List<Exit> exits = operations.peekExits(chainSpec.getMaxExits());
+    List<Deposit> deposits =
+        depositContract.peekDeposits(
+            chainSpec.getMaxDeposits(), state.getLatestEth1Data(), UInt64.ZERO);
```
Nashatyrev: Don't we need to specify the latest added `Deposit` here to peek starting from it ?
Did I get it right that `state.getLatestEth1Data()` is just the 'upper bound' for searching for new deposits?

mkalinin: Latest included deposit index should be a part of `BeaconState` which is not yet updated in the spec.
> Did I get it right that `state.getLatestEth1Data()` is just the 'upper bound' for searching for new deposits?

yes, exactly

Nashatyrev: Hm... I'm hoping they include this index as `Eth1Data` since searching just by index would require maintaining additional index

mkalinin: I see. Need a suggestion on keeping previous `Eth1Data` then.

chain/src/main/java/org/ethereum/beacon/chain/observer/PendingOperations.java
```diff
@@ -2,9 +2,20 @@
...
+import org.ethereum.beacon.core.operations.CasperSlashing;
+import org.ethereum.beacon.core.operations.Exit;
+import org.ethereum.beacon.core.operations.ProposerSlashing;
 
 /** A pending state interface. */
 public interface PendingOperations {
 
   List<Attestation> getAttestations();
+
+  List<ProposerSlashing> peekProposerSlashings(int count);
```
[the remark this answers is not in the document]
zilm13: > It's assumed to be specific to PendingOperations implementation.
Sounds reasonable!

PR #41, opened 2019-01-23
SSZ hasher added as hash_tree_root specHelper

zilm13:
+ one test for supported model

### DIFF
--- a/consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
+++ b/consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
@@ -55,14 +55,22 @@ public class SpecHelpers {
   private final ChainSpec spec;
   private final Hasher<Hash32> objectHasher;
 
+  /* Uses Hash32.ZERO as a stub for objectHasher */
   public SpecHelpers(ChainSpec spec) {
     this.spec = spec;
     this.objectHasher = input -> Hash32.ZERO;
   }
 
-  public SpecHelpers(ChainSpec spec, Function<Function<BytesValue, Hash32>, Hasher<Hash32>> objectHasherBuilder) {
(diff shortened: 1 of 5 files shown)


consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -45,12 +49,23 @@
...
 
 /**
  * https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#helper-functions
  */
 public class SpecHelpers {
   private final ChainSpec spec;
+  private static final Function<BytesValue, BytesValue> hashingFunction =
+      (data) ->
+          BytesValue.wrap(Hashes.keccak256(BytesValue.of(data.getArrayUnsafe())).getArrayUnsafe());
+  private static final SSZHasher sszHasher =
```
mkalinin: Better to avoid `static` members if it could be achieved.

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -45,12 +49,23 @@
...
+import java.util.function.Function;
 
 /**
  * https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#helper-functions
  */
 public class SpecHelpers {
   private final ChainSpec spec;
+  private static final Function<BytesValue, BytesValue> hashingFunction =
+      (data) ->
+          BytesValue.wrap(Hashes.keccak256(BytesValue.of(data.getArrayUnsafe())).getArrayUnsafe());
```
mkalinin: `Hashes.keccak256` to `this.hash`?

zilm13:
@mkalinin what do you think about this?

PR #46, opened 2019-01-25
Beacon attester

mkalinin:
### What's done
Beacon chain attester implementation in the first iteration.

Resolves https://github.com/harmony-dev/beacon-chain-java/issues/49


validator/src/main/java/org/ethereum/beacon/validator/attester/BeaconChainAttesterImpl.java
```diff
@@ -0,0 +1,127 @@
...
+        head.getSlot().minus(head.getSlot().modulo(chainSpec.getEpochLength()));
+    return specHelpers.get_block_root(state, epochBoundarySlot);
+  }
+
+  /*
+   Set attestation_data.latest_crosslink_root = state.latest_crosslinks[shard].shard_block_root
+     where state is the beacon state at head and shard is the validator's assigned shard.
+  */
+  private Hash32 getLatestCrosslinkRoot(BeaconState state, UInt64 shard) {
+    return state.getLatestCrosslinks().get(safeInt(shard)).getShardBlockRoot();
```
Nashatyrev: `safeInt` can be replaced with `UInt64.getIntValue()`. May be remove it from `SpecHelpers` ?

mkalinin: Created corresponding issue https://github.com/harmony-dev/beacon-chain-java/issues/50.

PR #51, opened 2019-01-30
Custom types

Nashatyrev:
Add custom types like `SlotNumber, ValidatorIndex`, etc. which were added to the spec. 
Also considering other types outside of the spec like `Timestamp, Bitfield`


core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.ethereum.beacon.ssz.annotation.SSZSerializable;
+import tech.pegasys.artemis.util.uint.UInt64;
+
+@SSZSerializable(serializeAs = UInt64.class)
+public class Bitfield extends UInt64 {
+
+  public static final Bitfield ZERO = new Bitfield(UInt64.ZERO);
+
+  public Bitfield(UInt64 uint) {
```
mkalinin: There are two bitfields in the domain. This one is `UInt64` based and used as a type for `justificationBitfield` field. Another one is based on `BytesValue` and has arbitrary length, used in `AttestationData`.

IMO, `Bitfield` name matches better for the latter case. As for `justificationBitfield` I really have no idea except `Bitfield64`.

core/src/main/java/org/ethereum/beacon/core/types/Time.java
```diff
@@ -0,0 +1,50 @@
+package org.ethereum.beacon.core.types;
+
+import org.ethereum.beacon.ssz.annotation.SSZSerializable;
+import tech.pegasys.artemis.util.uint.UInt64;
+
+@SSZSerializable(serializeAs = UInt64.class)
+public class Time extends UInt64 implements SafeComparable<Time> {
```
mkalinin: I think that `Timestamp` would more precise.

Nashatyrev: `Timestamp` was my initial effort. But this type refers both `timestamp` and `time period` (like `SlotDuration`), so I just named it `Time`

core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
zilm13: I've started to use `Bitfield` for participationBitfield, and it's definitely more sense to use it there

Nashatyrev: Sounds reasonable. Done!

core/src/main/java/org/ethereum/beacon/core/types/Time.java
Nashatyrev: Let's leave `Time` for now. I like `Seconds` less :/

Nashatyrev: Making separate `Timestamp` and `Period` types sounds like an overkill to me for existing 3 class usages

PR #56, opened 2019-02-01
Spec update to v0.1 (01/31/19)

Nashatyrev:
https://github.com/ethereum/eth2.0-specs/releases/tag/v0.1


consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -593,44 +725,30 @@ def activate_validator(state: BeaconState, index: int, genesis: bool) -> None:
       )
    */
   public void activate_validator(MutableBeaconState state, ValidatorIndex index, boolean genesis) {
-    SlotNumber activationSlot = genesis ?
-        spec.getGenesisSlot() :
-        state.getSlot().plus(spec.getEntryExitDelay());
+    EpochNumber activationSlot = genesis ?
```
zilm13: It's not in google style as I remember it

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -983,10 +1107,8 @@ def is_double_vote(attestation_data_1: AttestationData,
   */
   public boolean is_double_vote(
       AttestationData attestation_data_1, AttestationData attestation_data_2) {
-    EpochNumber target_epoch_1 = attestation_data_1.getSlot()
-        .dividedBy(spec.getEpochLength());
-    EpochNumber  target_epoch_2 = attestation_data_2.getSlot()
-        .dividedBy(spec.getEpochLength());
+    EpochNumber target_epoch_1 = slot_to_epoch(attestation_data_1.getSlot());
+    EpochNumber  target_epoch_2 = slot_to_epoch(attestation_data_2.getSlot());
```
zilm13:
```suggestion
    EpochNumber target_epoch_2 = slot_to_epoch(attestation_data_2.getSlot());
```

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1259,22 +1434,36 @@ public void checkShardRange(ShardNumber shard) {
...
+      attestation_targets = [
+          get_latest_attestation_target(store, validator)
+          for validator in active_validators
+      ]
+
+      def get_vote_count(block: BeaconBlock) -> int:
+          return len([
+              target
+              for target in attestation_targets
+              if get_ancestor(store, target, block.slot) == block
```
zilm13: `get_ancestor` was updated too from if-else to if-elif-else:
```
Let get_ancestor(store: Store, block: BeaconBlock, slot: SlotNumber) -> BeaconBlock be the ancestor of block with slot number slot. The get_ancestor function can be defined recursively as:
def get_ancestor(store: Store, block: BeaconBlock, slot: SlotNumber) -> BeaconBlock:
    """
    Get the ancestor of ``block`` with slot number ``slot``; return ``None`` if not found.
    """
    if block.slot == slot:
        return block
    elif block.slot < slot:
        return None
    else:
        return get_ancestor(store, store.get_parent(block), slot)
```
proof https://github.com/ethereum/eth2.0-specs/blob/4cdb66728a27739fcb35e545386770f3a2972bf8/specs/core/0_beacon-chain.md

Nashatyrev: Good catch!

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
Nashatyrev: What exactly violated here?

zilm13:
```suggestion
       EpochNumber activationSlot =
        genesis ? spec.getGenesisEpoch() : get_entry_exit_effect_epoch(get_current_epoch(state));
```
or 
```suggestion
        EpochNumber activationSlot = genesis 
        ? spec.getGenesisEpoch() 
        : get_entry_exit_effect_epoch(get_current_epoch(state));
```

by default it suggests first as it fits in line. If you make it longer it will suggest 2nd

PR #60, opened 2019-02-05
Make slot ticker a bit more robust

mkalinin:
A slight improvement of delay and slot management in a `SlotTicker`.

### DIFF
--- a/chain/src/main/java/org/ethereum/beacon/chain/SlotTicker.java
+++ b/chain/src/main/java/org/ethereum/beacon/chain/SlotTicker.java
@@ -7,6 +7,7 @@ import org.ethereum.beacon.consensus.SpecHelpers;
 import org.ethereum.beacon.core.BeaconState;
 import org.ethereum.beacon.core.types.SlotNumber;
 import org.ethereum.beacon.core.types.Time;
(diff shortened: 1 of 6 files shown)


chain/src/main/java/org/ethereum/beacon/chain/SlotTicker.java
```diff
@@ -31,35 +30,41 @@
...
+    Time period = specHelpers.getChainSpec().getSlotDuration();
+
+    startImpl(nextSlot, period, Schedulers.fromExecutor(scheduler));
+  }
+
+  private void startImpl(SlotNumber startSlot, Time period, Scheduler scheduler) {
+    this.startSlot = startSlot;
+
+    Time startSlotTime = specHelpers.get_slot_start_time(state, startSlot);
+    long delayMillis = Math.max(0, startSlotTime.getMillis() - System.currentTimeMillis());
```
zilm13: please, use one time source in code
our application is time critical and will need to use one accurate source of time

mkalinin: makes a lot of sense to me 👍

PR #61, opened 2019-02-06
Tests for validator service

mkalinin:
### What's done
Written tests covers:
- BeaconChainProposerImpl
- BeaconChainAttesterImpl
- BeaconChainValidator

Also, various utility classes has been created to facilitate writing of the tests in the future. These utilities are sited in corresponding modules, e.g. `AttestationTestUtil` is inside `core` module. The idea behind this placement is simple, when module is specified as dependency one gets these utilities out of the box. 

But gradle doesn't import test sources by default. And a hack is required to make it work: https://github.com/harmony-dev/beacon-chain-java/blob/test/validator-service/validator/build.gradle#L15-L18. There is a number of solutions to this problem, but any of them is a hack.

Please, pay attention to this utility classes and their usage. If we don't want to have anything like that or there are any ideas on a better approach, feel free to let me know.


chain/src/test/java/org/ethereum/beacon/chain/util/PendingOperationsTestUtil.java
```diff
@@ -0,0 +1,39 @@
...
+  public static PendingOperations mockPendingOperations(
+      List<Attestation> attestations,
+      List<ProposerSlashing> proposerSlashings,
+      List<CasperSlashing> casperSlashings,
+      List<Exit> exits) {
+    PendingOperations pendingOperations = Mockito.mock(PendingOperations.class);
+    when(pendingOperations.getAttestations()).thenReturn(attestations);
+    when(pendingOperations.peekProposerSlashings(anyInt())).thenReturn(proposerSlashings);
+    when(pendingOperations.peekCasperSlashings(anyInt())).thenReturn(casperSlashings);
+    when(pendingOperations.peekAggregatedAttestations(anyInt(), any())).thenReturn(attestations);
```
zilm13: I'd split attestations and aggregated attestations in this method and init it in `createEmptyPendingOperations` with the same list if you want

pow/core/build.gradle
```diff
@@ -4,4 +4,9 @@ dependencies {
   implementation project(':ssz')
 
   implementation 'io.projectreactor:reactor-core'
+  implementation 'com.google.guava:guava'
+
+  testImplementation 'org.mockito:mockito-core'
+
+  testImplementation project(':core').sourceSets.test.output
```
zilm13: Is it some kind of hack?
If yes, I'd add comment

types/src/main/java/tech/pegasys/artemis/util/bytes/Bytes48.java
```diff
@@ -178,6 +179,19 @@ static Bytes48 fromHexStringStrict(String str) {
     return wrap(BytesValues.fromRawHexString(str, -1, false));
   }
 
+  /**
+   * Constructs a randomly generated value.
+   *
+   * @param rnd random number generator.
+   * @return random value.
+   */
+  static Bytes48 random(Random rnd) {
```
zilm13: Why not public?
```suggestion
  public static Bytes48 random(Random rnd) {
```

types/src/main/java/tech/pegasys/artemis/util/bytes/Bytes96.java
```diff
@@ -125,6 +127,18 @@ static Bytes96 fromHexStringStrict(String str) {
     return wrap(BytesValues.fromRawHexString(str, -1, false));
   }
 
+  /**
+   * Constructs a randomly generated value.
+   *
+   * @param rnd random number generator.
+   * @return random value.
+   */
+  static Bytes96 random(Random rnd) {
```
zilm13:
```suggestion
  public static Bytes96 random(Random rnd) {
```

validator/src/main/java/org/ethereum/beacon/validator/BeaconChainValidator.java
```diff
@@ -46,11 +44,11 @@
   private MessageSigner<BLSSignature> messageSigner;
 
   /** Validator index. Assigned in {@link #init(BeaconState)} method. */
-  private ValidatorIndex validatorIndex = ValidatorIndex.MAX;
+  @VisibleForTesting ValidatorIndex validatorIndex = ValidatorIndex.MAX;
```
zilm13: I don't like this thing
We have java's private/package-only/protected/public and this thing breaks the things forcing it package-only instead of private
@Nashatyrev  what do you think?

types/src/main/java/tech/pegasys/artemis/util/bytes/Bytes48.java
mkalinin: Cause it's an interface.

pow/core/build.gradle
mkalinin: Yep, it's naturally a hack.

types/src/main/java/tech/pegasys/artemis/util/bytes/Bytes48.java
zilm13: Ohh my bad

validator/src/main/java/org/ethereum/beacon/validator/BeaconChainValidator.java
mkalinin: Yep, this is something I don't like either. But sometimes for testing this is the only way to run checks easily, without additional tweaks.

Nashatyrev: Yep, looks pretty crappy. 
It's better than declare package-private and put tests to the same package.

PR #62, opened 2019-02-07
Get rid of warnings in SSZ tests

mkalinin:
### What's done
Gets rid of warnings in SSZ tests by replacing Hex API from Sun to `BytesValue.fromHexString()`/`BytesValue.toString()`.


zilm13 (approved):
Wonderful.

Issue #45, opened 2019-01-25
Add time frames block verification


mkalinin:
### What's wrong
There is no sanity check for a slot time frames during block import.

Issue #47, opened 2019-01-28
Fallback to disk storage when fetching justfified blocks


mkalinin:
### What's wrong
Currently, justified block hashes are fetched via [SpecHelpers.get_block_root](https://github.com/harmony-dev/beacon-chain-java/blob/master/consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java#L1056) method which is limited with `8,192` blocks by [LATEST_BLOCK_ROOTS_LENGTH](https://github.com/harmony-dev/beacon-chain-java/blob/master/core/src/main/java/org/ethereum/beacon/core/spec/MiscParameters.java#L22) parameter. Hence, there is potential assertion error in there.

### Solution
Address query to disk storage when distance to a slot of justified block is greater than `LATEST_BLOCK_ROOTS_LENGTH`.

mkalinin:
Addressed by https://github.com/harmony-dev/beacon-chain-java/pull/65

mkalinin:
This is still actual for [BeaconChainAttesterImpl.getJustifiedBlockRoot()](https://github.com/harmony-dev/beacon-chain-java/blob/feature/emulator-1/validator/src/main/java/org/ethereum/beacon/validator/attester/BeaconChainAttesterImpl.java#L136-L137)

PR #65, opened 2019-02-07
Beacon chain updates

mkalinin:
### What's done
Updates `DefaultBeaconChain` to latest spec and design changes, featured with a number of refactor updates.

Resolves https://github.com/harmony-dev/beacon-chain-java/issues/45 https://github.com/harmony-dev/beacon-chain-java/issues/47


core/src/main/java/org/ethereum/beacon/core/Hashable.java
```diff
@@ -3,12 +3,4 @@
...
-public interface Hashable<H extends Hash> {
-
-  /**
-   * Returns a hash of the object.
-   *
-   * @return a hash.
-   */
-  H getHash();
-}
+public interface Hashable<H extends Hash> {}
```
zilm13: Why do we need it now?

mkalinin: I left it as a part of type safety but there is not much benefit in that.

PR #70, opened 2019-02-10
Validator registration service

zilm13:
What should be done:
Service which submits and handles data from inputs to active validator in ethereum 2.0
so:
1. Create and submit deposit
2. Wait until validator is active for proposing/attesting
3. Run attester/proposer service

Service should be able to resume correct stage after restart.


zilm13:
I think I'm done, one thing to done: persist current registration stage to DB. Do we have some way to store any variable?

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/ValidatorRegistrationServiceImpl.java
```diff
@@ -0,0 +1,298 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Function;
+
+import static org.ethereum.beacon.core.spec.SignatureDomains.DEPOSIT;
+
+public class ValidatorRegistrationServiceImpl implements ValidatorRegistrationService {
```
mkalinin: Maybe we don't need a complete service for that? I mean that it's not always necessary to run a full flow here. I think we can keep it as it is now and then break down to the components that would be really used.

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/ValidatorRegistrationServiceImpl.java
```diff
@@ -0,0 +1,298 @@
...
+    getCurrentStageFunction(currentStage).run();
+  }
+
+  private void changeCurrentStage(RegistrationStage newStage) {
+    this.currentStage = newStage;
+    stagePersistence.set(currentStage);
+    actionCircle();
+  }
+
+  private RegistrationStage getCurrentStage() {
```
mkalinin: This method doesn't do a check if this validator has already been registered on contract. It might be a useful thing.

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/TransactionBuilder.java
```diff
@@ -0,0 +1,13 @@
+package org.ethereum.beacon.pow.validator;
+
+import org.ethereum.beacon.core.types.Gwei;
+import tech.pegasys.artemis.util.bytes.BytesValue;
+
+/** Util for creating validator transactions */
+public interface TransactionBuilder {
+  boolean isReady();
```
mkalinin: Do we need an explicit check here? I think that a future which completes when sync is done would be sufficient enough for all interactions with Eth1. IMO, It should be handle on the higher abstraction layer than this.

pow/ethereumj/src/main/java/org/ethereum/beacon/pow/EthereumJTransactionBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+    BigInteger nonce =
+        ethereum.getRepository().getNonce(Address.fromHexString(fromAddress).getArrayUnsafe());
+    byte[] data = contract.getByName("deposit").encode((Object) depositInput.getArrayUnsafe());
+    Transaction tx =
+        ethereum.createTransaction(
+            nonce,
+            BigInteger.valueOf(gasPriceTracker.getRecommendedGasPrice()),
+            BigInteger.valueOf(2_000_000), // FIXME: Why this number??
+            contractDeployAddress.getArrayUnsafe(),
+            amount.weiValue(),
```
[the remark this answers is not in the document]
zilm13: I thought there are some sense to have it in our class as we are going to use it not only with EthereumJ

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/ValidatorRegistrationServiceImpl.java
zilm13: Not sure at this time. Will see it better when web3j or go added

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/ValidatorRegistrationServiceImpl.java
```diff
@@ -0,0 +1,298 @@
...
+      Publisher<ObservableBeaconState> observablePublisher,
+      SingleValueSource<RegistrationStage> registrationStagePersistence,
+      SpecHelpers specHelpers) {
+    this.transactionBuilder = transactionBuilder;
+    this.transactionGateway = transactionGateway;
+    this.depositContract = depositContract;
+    this.observablePublisher = observablePublisher;
+    this.stagePersistence = registrationStagePersistence;
+    this.specHelpers = specHelpers;
+    Function<BytesValue, Hash32> hashFunction = Hashes::keccak256;
```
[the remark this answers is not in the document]
zilm13: Actually I could get it from specHelpers

pow/validator/src/main/java/org/ethereum/beacon/pow/validator/ValidatorRegistrationServiceImpl.java
zilm13: How about the approach I made?

pow/core/src/main/java/org/ethereum/beacon/pow/AbstractDepositContract.java
```diff
@@ -67,6 +78,7 @@ protected synchronized void chainStart(byte[] deposit_root, byte[] time, byte[]
             Hash32.wrap(Bytes32.wrap(blockHash))),
         initialDeposits);
     chainStartSink.onNext(chainStart);
+    initialDeposits.forEach(depositSink::onNext);
```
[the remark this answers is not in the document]
zilm13: My fail, fixed

pow/ethereumj/src/main/java/org/ethereum/beacon/pow/EthereumJTransactionBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+                  .getNonce(Address.fromHexString(fromAddress).getArrayUnsafe());
+          byte[] data =
+              contract.getByName("deposit").encode((Object) depositInput.getArrayUnsafe());
+          Transaction tx =
+              ethereum.createTransaction(
+                  nonce,
+                  BigInteger.valueOf(gasPriceTracker.getRecommendedGasPrice()),
+                  BigInteger.valueOf(2_000_000), // FIXME: Why this number??
+                  contractDeployAddress.getArrayUnsafe(),
+                  amount.weiValue(),
```
mkalinin: Anyway, converting value is an implementation specific thing. Look here https://web3js.readthedocs.io/en/1.0/web3-utils.html#towei

PR #71, opened 2019-02-12
Fixes to existing infrastructure to get emulator up and running 

Nashatyrev:
This is the first phase of creating emulator: several validator services which interact through a local Wire stub. 
In this phase I will be collecting fixes and refactorings to existing classes


chain/src/test/java/org/ethereum/beacon/chain/StreamTests.java
```diff
@@ -43,4 +42,25 @@ public void test1() throws InterruptedException {
 
     Thread.sleep(1000L);
   }
+
+  @Test
+  @Ignore
+  public void intervalTest1() throws InterruptedException {
```
zilm13: What are you testing here?

chain/src/main/java/org/ethereum/beacon/chain/observer/ObservableStateProcessorImpl.java
```diff
@@ -53,32 +52,35 @@
   private final Publisher<BeaconTuple> beaconPublisher;
 
   private static final int UPDATE_MILLIS = 500;
-  private ScheduledExecutorService regularJobExecutor;
-  private ScheduledExecutorService continuousJobExecutor;
+  private Scheduler regularJobExecutor =
```
zilm13: is it lazy? we don't need both until we need this service started

chain/src/main/java/org/ethereum/beacon/chain/observer/ObservableStateProcessorImpl.java
```diff
@@ -103,29 +105,22 @@ public ObservableStateProcessorImpl(
...
+    Flux.from(slotTicker).subscribe(this::onNewSlot);
+    Flux.from(attestationPublisher).subscribe(this::onNewAttestation);
+    Flux.from(beaconPublisher).subscribe(this::onNewBlockTuple);
+    regularJobExecutor.executeAtFixedRate(
+        Duration.ZERO, Duration.ofMillis(UPDATE_MILLIS), this::doHardWork);
   }
 
   private void runTaskInSeparateThread(Runnable task) {
-    continuousJobExecutor.execute(task);
+    continuousJobExecutor.execute(() -> {
```
zilm13: maybe add `execute(Runnable task, Consumer<Exception> failCallback)`?

validator/src/main/java/org/ethereum/beacon/validator/BeaconChainValidator.java
```diff
@@ -294,11 +307,17 @@ private boolean isInitialized() {
   }
 
   /* FIXME: stub for streams. */
```
mkalinin: Let's remove this comment as it looks outdated.

chain/src/test/java/org/ethereum/beacon/chain/StreamTests.java
Nashatyrev: Just some experiments. Actually `SlotTicker` should consist just of several lines, but `Flux.interval` has missing feature: it accounts `initialDelay` only on subscribing. Will be working on this

chain/src/main/java/org/ethereum/beacon/chain/observer/ObservableStateProcessorImpl.java
Nashatyrev: This is just a temp solution. You may always handle returned `CompletableFuture`, but I'm going to make generic solution for all executors to log unhandled exceptions

chain/src/main/java/org/ethereum/beacon/chain/observer/ObservableStateProcessorImpl.java
Nashatyrev: Nope :/ 
Moved them back to `start()`

PR #79, opened 2019-02-20
Get rid of Schedulers.get() statics

Nashatyrev:
Pass `Schedulers` to all classes as parameter. 
This will help to separate validator instances running in a single jvm emulator. This will also allow to manage separate logging contexts for different validators

### DIFF
--- a/chain/src/test/java/org/ethereum/beacon/chain/DefaultBeaconChainTest.java
+++ b/chain/src/test/java/org/ethereum/beacon/chain/DefaultBeaconChainTest.java
@@ -17,10 +17,12 @@ import org.ethereum.beacon.consensus.verifier.VerificationResult;
 import org.ethereum.beacon.core.BeaconBlock;
 import org.ethereum.beacon.core.BeaconBlockBody;
 import org.ethereum.beacon.core.BeaconState;
+import org.ethereum.beacon.core.spec.ChainSpec;
 import org.ethereum.beacon.core.state.Eth1Data;
 import org.ethereum.beacon.core.types.Time;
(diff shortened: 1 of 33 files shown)


Nashatyrev:
@zilm13 Having a separate `TimeSupplier` makes sense only for `ControlledSchedulers` but would be complex and impractical for `DefaultSchedulers`

chain/src/test/java/org/ethereum/beacon/chain/DefaultBeaconChainTest.java
```diff
@@ -30,10 +33,12 @@
 
   @Test
   public void insertAChain() {
-    SpecHelpers specHelpers = SpecHelpers.createDefault();
+    Schedulers schedulers = new DefaultSchedulers();
```
mkalinin: What if we keep `Schedulers.getDefault()` to instantiate a `DefaultSchedulers` object? First, this is convenient, second, it would be easier to swap a default implementation of `Schedulers` later on.

mkalinin: Third, it increases UX of `Schedulers`, user would be able to pick some default implementation without additional investigation.

Nashatyrev: Yep, I was going to extract `Schedulers` interface a bit later, but let's do it in this PR...

mkalinin: **EDT:** `Schedulers.createDefault()` is better.

PR #83, opened 2019-02-26
Multi-validator service

mkalinin:
### What's done
A service that able to carry `N` validators in one instance. 
An implementation is simply a copy&paste of `BeaconChainValidator` code adopted for a multi-validator case. As these two classes has a lot in common, it could be moved to some `AbstractValidatorService` later on, if needed.

### DIFF
--- a/validator/src/main/java/org/ethereum/beacon/validator/MultiValidatorService.java
+++ b/validator/src/main/java/org/ethereum/beacon/validator/MultiValidatorService.java
@@ -102,7 +102,7 @@ public class MultiValidatorService implements ValidatorService {
             .name("BeaconChainValidator.attestation");
 
     executor = this.schedulers.newSingleThreadDaemon("validator-service");
-    initExecutor = new LatestExecutor<>(schedulers.cpuHeavy(), this::initFromLatestBeaconState);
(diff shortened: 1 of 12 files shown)


Nashatyrev:
We need to use only `Scheduler` instances across the project. I've added specific `LatestExecutor` class for this (I also needed it `pow` module and will replace later)

validator/src/main/java/org/ethereum/beacon/validator/MultiValidatorService.java
```diff
@@ -0,0 +1,375 @@
...
+  public void stop() {}
+
+  /**
+   * Passes validators from {@link #uninitialized} map through initialization process.
+   *
+   * @param state a state object to seek for validators in.
+   */
+  private void initFromLatestBeaconState(BeaconState state) {
+    Map<ValidatorIndex, BLS381Credentials> intoCommittees = new HashMap<>();
+    for (ValidatorIndex i : state.getValidatorRegistry().size()) {
```
Nashatyrev: Potential place for optimization: it would be too heavy to iterate over all validators each new `BeaconState`.

PR #84, opened 2019-02-26
Emulator further bugfixes and improvements 

Nashatyrev:
This is continue of work to make virtual local net of several peers functioning correctly.
- Several bugs in spec implementation fixed.
- Improved logging, toString(...) info
- Fix the proposer to build block on the state without epoch transition
- Fix `ObservableStateProcessor` to keep all the attestations from a single verifier


chain/src/main/java/org/ethereum/beacon/chain/observer/ObservableBeaconState.java
```diff
@@ -10,12 +10,23 @@
 public class ObservableBeaconState {
   private final BeaconBlock head;
   private final BeaconState latestSlotState;
+  private final BeaconState latestSlotStateWithoutEpoch;
```
zilm13: Could you add some description anyway?
Still not clear for me from the name.

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1265,9 +1265,13 @@ boolean verify_bitfield(Bitfield bitfield, int committee_size) {
       return false;
     }
 
-    for(int i = committee_size + 1; i < committee_size - committee_size % 8 + 8; i++) {
-      if (bitfield.getBit(i)) {
-        return false;
+    for(int i = committee_size; i < bitfield.size() * 8; i++) {
```
zilm13: space after `for`, please

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1265,9 +1265,13 @@ boolean verify_bitfield(Bitfield bitfield, int committee_size) {
...
+      try {
+        if (bitfield.getBit(i)) {
+          return false;
+        }
+      } catch (Exception e) {
+        e.printStackTrace();
```
zilm13: maybe rethrow?

PR #94, opened 2019-02-28
start/cli: use literal log levels instead of numbers

zilm13:


start/cli/src/main/java/org/ethereum/beacon/ReusableOptions.java
```diff
@@ -22,17 +22,17 @@
...
   static {
-    LOG_LEVELS.put(0, Level.OFF);
-    LOG_LEVELS.put(1, Level.FATAL);
-    LOG_LEVELS.put(2, Level.ERROR);
-    LOG_LEVELS.put(3, Level.WARN);
-    LOG_LEVELS.put(4, Level.INFO);
-    LOG_LEVELS.put(5, Level.DEBUG);
-    LOG_LEVELS.put(6, Level.TRACE);
-    LOG_LEVELS.put(7, Level.ALL);
+    LOG_LEVELS.put(LogLevel.OFF, Level.OFF);
```
mkalinin: Do we really need all of them? I am not sure if our code even uses them all. What if we keep those which could be really useful?

PR #95, opened 2019-02-28
Ssz updates

### LINKED ISSUE #19: Switch integer encoding in SSZ to little endian
If pull request is merged 
https://github.com/ethereum/eth2.0-specs/pull/139
we should change our implementation to handle numbers as little endians

zilm13:
Fixes all ssz debts:
1) resolves #19
2) renames `hash_tree_root` to `hash_tree_root_internal`
[ethereum/eth2.0-specs#543](https://github.com/ethereum/eth2.0-specs/pull/543)
3) uses `merkle_hash` instead of `hash` for containers
[ethereum/eth2.0-specs#595](https://github.com/ethereum/eth2.0-specs/pull/595)
it had error in the spec but spec is going to be rewritten with fix of this error, so it's not up to v0.4 version spec, because it has my own fix and it's not up to rewritten spec, because it's not final. Anyway 0.4 is not correct in ssz
4) Tree hash `bytes[]` as `List(Bytes1)` and skip length prefix for fixed byte arrays
[ethereum/eth2.0-specs#596](https://github.com/ethereum/eth2.0-specs/pull/596)

### DIFF
--- a/consensus/src/main/java/org/ethereum/beacon/consensus/hasher/SSZObjectHasher.java
+++ b/consensus/src/main/java/org/ethereum/beacon/consensus/hasher/SSZObjectHasher.java
@@ -19,26 +19,20 @@ import java.util.function.Function;
 public class SSZObjectHasher implements ObjectHasher<Hash32> {
 
   private final SSZHashSerializer sszHashSerializer;
-  private final Function<BytesValue, Hash32> hashFunction;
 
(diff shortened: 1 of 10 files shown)


zilm13:
Please, don't merge
Will try to update according to rewritten hash_tree spec, if it changes anything or confirm that this implementation matches it.

zilm13:
5. implemented spec rewrite https://github.com/ethereum/eth2.0-specs/pull/696
still WIP, need to clarify several things

PR #97, opened 2019-03-04
Fixes added when profiling (optimization)

zilm13:
+ cache in ssz scheme builder


zilm13:
Added caches:
1.Unlimited for SSZ classes scheme builder, was called on any serialization/hashing
2. Limited with removing oldest entry for `hash_tree_root` and `signed_root`, because it's often called on the same objects without modifications, for example for logging purposes

Plus modified list copy from streams collector to new arraylist, because of benchmarking:
```
old.stream().collect()
execution time: IntSummaryStatistics{count=1000, sum=167881796, min=56123, average=167881.796000, max=64305416}
new ArrayList(old)
execution time: IntSummaryStatistics{count=1000, sum=24309018, min=18468, average=24309.018000, max=80586}
new ArrayList().addAll(old)
execution time: IntSummaryStatistics{count=1000, sum=44968500, min=9780, average=44968.500000, max=3953362}
```

consensus/src/main/java/org/ethereum/beacon/consensus/SpecHelpers.java
```diff
@@ -1002,12 +1009,15 @@ def eligible(index):
 
   /** Function for hashing objects into a single root utilizing a hash tree structure */
   public Hash32 hash_tree_root(Object object) {
-    return objectHasher.getHash(object);
+    String id = object.getClass().getName() + "@" + Integer.toHexString(object.hashCode());
```
Nashatyrev: I think `System.identityHashCode(object)` would be more suitable here

zilm13: outdate

core/src/main/java/org/ethereum/beacon/core/types/Hash32able.java
```diff
@@ -0,0 +1,14 @@
+package org.ethereum.beacon.core.types;
+
+import tech.pegasys.artemis.ethereum.core.Hash32;
+
+import java.util.Optional;
+
+/** Indicates type with {@link Hash32} hash */
+public interface Hash32able {
```
mkalinin: Maybe it could be a generic interface like `Hashable<HASH extends BytesValue>`?

zilm13:
ready for review and merge

PR #98, opened 2019-03-05
Update consensus up to spec version 0.4.0

mkalinin:
### What's done
Consensus module with all the references is updated to version [0.4.0](https://github.com/ethereum/eth2.0-specs/releases/tag/0.4.0) of the spec.


consensus/src/main/java/org/ethereum/beacon/consensus/verifier/block/TransferListVerifier.java
```diff
@@ -16,6 +19,19 @@ protected TransferListVerifier(OperationVerifier<Transfer> operationVerifier, Sp
         operationVerifier,
         block -> block.getBody().getTransfers(),
         spec.getConstants().getMaxTransfers());
+
+    addCustomVerifier(
+        ((transfers, state) -> {
+          for (Transfer t1 : transfers) {
+            for (Transfer t2 : transfers) {
+              if (t1.equals(t2)) {
+                return failedResult("two equal transfers have been found");
```
Nashatyrev: Hm, shouldn't that immediately fail on the first iteration when `t1 == transfers[0], t2 == transfers[0]` ?

Nashatyrev: M.b. something like this:
```java
if (new HashSet(transfers).size() != transfers.size()) // fail
```

mkalinin: hah! nice catch

PR #102, opened 2019-03-09
CLI imrovements

mkalinin:
### What's done
Work on CLI improvement is finished. It's been simplified in UI part and in the code as well.

CLI samples:
```
./simulator
Usage: simulator [-hV] [COMMAND]
Beacon chain simulator
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  run   Runs simulation
  spec  Prints beacon chain constants defined by the spec
```

```
./simulator run
Missing required parameter: plan
Usage: simulator run [-hV] [--loglevel=level] [--spec=spec.yml] plan [COMMAND]
Runs simulation
      plan               A path to a simulation plan in YAML format
      --loglevel=level   Log verbosity level: all, debug, info, error
                         info is set by default
      --spec=spec.yml    A path to YAML file with beacon chain constants
                         values from given file override values from the spec
  -h, --help             Show this help message and exit.
  -V, --version          Print version information and exit.
Commands:
  default  Runs simulation with default setup
```

To start simulation with default config and spec constants use:
```
./simulator run default
```


start/cli/src/main/java/org/ethereum/beacon/Simulator.java
```diff
@@ -1,87 +1,38 @@
...
+  private static final int ERROR_EXIT_CODE = 1;
 
   public static void main(String[] args) {
     try {
-      CommandLine.call(new Simulator(), args);
+      CommandLine commandLine = new CommandLine(new Simulator());
+      commandLine.setCaseInsensitiveEnumValuesAllowed(true);
+      commandLine.parseWithHandlers(
+          new RunLast().andExit(SUCCESS_EXIT_CODE),
+          CommandLine.defaultExceptionHandler().andExit(ERROR_EXIT_CODE),
```
zilm13: :+1:

start/cli/src/main/java/org/ethereum/beacon/command/RunDefaultSimulation.java
```diff
@@ -0,0 +1,40 @@
...
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "default",
+    description = "Runs simulation with default setup",
+    mixinStandardHelpOptions = true)
+public class RunDefaultSimulation implements Runnable {
+
+  @CommandLine.Option(
+      names = {"--loglevel"},
```
zilm13: What's the sense in `LogLevelAware` defined above this way?

start/cli/src/main/java/org/ethereum/beacon/command/PrintSpec.java
```diff
@@ -0,0 +1,21 @@
...
+    name = "spec",
+    description = "Prints beacon chain constants defined by the spec")
+public class PrintSpec implements Runnable {
+
+  @Override
+  public void run() {
+    Spec spec =
+        new ConfigBuilder<>(Spec.class)
+            .addYamlConfigFromResources("/config/spec-constants.yml")
+            .build();
```
zilm13: I don't like that it doesn't print current task merged spec. Do you think it's better for user to merge it he-self?

start/cli/src/main/java/org/ethereum/beacon/command/RunSimulation.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @CommandLine.Option(
+      names = {"--spec"},
+      paramLabel = "spec.yml",
+      description =
+          "A path to YAML file with beacon chain constants\nvalues from given file override values from the spec")
+  private File specOverrides;
+
+  @CommandLine.Option(
+      names = {"--loglevel"},
```
zilm13: ++ for `LogLevelAware`

start/cli/src/main/java/org/ethereum/beacon/command/RunSimulation.java
```diff
@@ -0,0 +1,54 @@
...
+  public void run() {
+    SimulatorConfig simulationPlan =
+        new ConfigBuilder<>(SimulatorConfig.class).addYamlConfig(plan).build();
+
+    ConfigBuilder<Spec> specBuilder =
+        new ConfigBuilder<>(Spec.class).addYamlConfigFromResources("/config/spec-constants.yml");
+    if (specOverrides != null) {
+      specBuilder.addYamlConfig(specOverrides);
+    }
+    Spec spec = specBuilder.build();
```
zilm13: One thing why I don't like we doesn't have plan print - because user doesn't know what's the options are available

mkalinin:
@zilm13 
> so, if I use keyword default, why do I need plan? Also, if default is the only option, why to list it like one of option.

Would it be better this way?
```
./simulator run
Missing required parameter: plan
Usage: simulator run [-hV] [--loglevel=level] [--spec=spec.yml] plan [COMMAND]
Runs simulation
      plan               A path to a simulation plan in YAML format
                         pass 'default' to run a simulation with default plan
      --loglevel=level   Log verbosity level: all, debug, info, error
                         info is set by default
      --spec=spec.yml    A path to YAML file with beacon chain constants
                         values from given file override values from the spec
  -h, --help             Show this help message and exit.
  -V, --version          Print version information and exit.
```
Then `--spec` option should not work for `default` plan (as default plan is tightly coupled with constants setup). Would it be less ambiguous?

start/cli/src/main/java/org/ethereum/beacon/command/RunDefaultSimulation.java
mkalinin: The intention was to share an option between two commands. But inheritance is not the best way of sharing a single option.

start/cli/src/main/java/org/ethereum/beacon/command/PrintSpec.java
mkalinin: In my opinion, if user sets up it's own constants he should be aware of those values. Btw, what do you think about printing all overridden spec constants right after the plan when simulation begins?

start/cli/src/main/java/org/ethereum/beacon/command/RunSimulation.java
mkalinin: The plan is printed at the beginning of each simulation. And we're going to have a simulator guide that would describe all the options available to user.

start/cli/src/main/java/org/ethereum/beacon/command/RunDefaultSimulation.java
zilm13: Did you check [mixins](https://picocli.info/#_mixins) ?
If both inheritance and mixins doesn't look good for you, just remove `LogLevelAware`, let it go!

start/cli/src/main/java/org/ethereum/beacon/command/PrintSpec.java
zilm13: > printing all overridden spec constants
cool!

start/cli/src/main/java/org/ethereum/beacon/command/RunSimulation.java
zilm13: Good, it looks fine for current stage.

start/cli/src/main/java/org/ethereum/beacon/command/RunDefaultSimulation.java
mkalinin: ahaha

zilm13:
` plan [COMMAND]`
so it will be `plan default` or `plan /my/path/to/the/plan.yml`?
If yes, still confusing, at least because of `COMMAND` sign when main purpose is path

mkalinin:
New run command interface:
```
simulator run
Missing required parameter: plan
Usage: simulator run [-hV] [--loglevel=level] [--spec=spec.yml] plan
Runs simulation
      plan               A path to a simulation plan in YAML format
                         pass 'default' to run a simulation with default setup
      --loglevel=level   Log verbosity level: all, debug, info, error
                         info is set by default
      --spec=spec.yml    A path to YAML file with beacon chain constants
                         values from given file override values from the spec
  -h, --help             Show this help message and exit.
  -V, --version          Print version information and exit.```

**Comment by mkalinin** [comment]
Beginning of the simulation now looks like this:
```./simulator run default
14:23:45.046 [main] INFO  # simulator - Simulation parameters:
---
peers:
- count: 8
  validator: true
- count: 1
  validator: false
seed: 1
bls-verify: false
genesis-time: 600

14:23:45.057 [main] INFO  # simulator - Overridden beacon chain parameters:
---
specConstants:
  honestValidatorParameters:
    ETH1_FOLLOW_DISTANCE: 1
  initialValues:
    GENESIS_SLOT: "1000000"
  miscParameters:
    shuffleRoundCount: 90
    SHARD_COUNT: "4"
    TARGET_COMMITTEE_SIZE: "2"
  stateListLengths:
    LATEST_BLOCK_ROOTS_LENGTH: "64"
    LATEST_RANDAO_MIXES_LENGTH: "64"
    LATEST_ACTIVE_INDEX_ROOTS_LENGTH: "64"
    LATEST_SLASHED_EXIT_LENGTH: "64"
  timeParameters:
    SECONDS_PER_SLOT: "10"
    MIN_ATTESTATION_INCLUSION_DELAY: "1"
    SLOTS_PER_EPOCH: "4"

14:23:45.525 [main] INFO  # simulator - Creating validators...
```

PR #108, opened 2019-03-15
Some config refactor

Nashatyrev:
Goals:
- move ChainSpec config (overridden properties) to the main config for conveniency
- return `blsVerify` option back to `SpecHelpersOptions` since it is rather Spec than simulation property


start/config/src/main/java/org/ethereum/beacon/emulator/config/main/plan/SimulationPlan.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+import org.ethereum.beacon.emulator.config.Config;
+import org.ethereum.beacon.emulator.config.YamlPrinter;
+import org.ethereum.beacon.emulator.config.simulator.PeersConfig;
+
+public class SimulationPlan extends Plan implements Config {
+
+  private List<PeersConfig> peers = new ArrayList<>();
+
+  @JsonProperty("genesis-time")
```
zilm13: why to keep this one dashed?

Nashatyrev: Thanks! Fixed

PR #112, opened 2019-03-29
State tests

zilm13:
Implements state test suite and enables it for unit tests.

TODO: revisit tests when `signing_root` and `hash_tree_root` fixed, https://github.com/harmony-dev/beacon-chain-java/issues/116.


consensus/src/main/java/org/ethereum/beacon/consensus/spec/BlockProcessing.java
```diff
@@ -30,7 +30,8 @@ default void process_block_header(MutableBeaconState state, BeaconBlock block) {
     // Verify that the slots match
     assertTrue(block.getSlot().equals(state.getSlot()));
     // Verify that the parent matches
-    assertTrue(block.getPreviousBlockRoot().equals(signed_root(state.getLatestBlockHeader())));
+    // FIXME: signed_root should match
```
zilm13: Let's not forget this thing

mkalinin: I'd better create an issue and point this out in the issue.

PR #118, opened 2019-04-14
Hash tree root github tests. 

zilm13:
Composite tests not finished, refactoring needed for all tests
Tests not included in repository, until this should be downloaded from https://gist.github.com/jannikluhn/e1440053bd53c16e04f06222b2e311e9 and copied to directory `beacon-chain-java/test/src/test/resources/eth2.0-tests/tree_hash` manually.


zilm13:
`@Ignore` lines could be removed when https://github.com/ethereum/eth2.0-test-generators/pull/34 is merged and tests are generated

zilm13:
TODO: composite test should be finished

zilm13:
Composite test removed, because it's too complex and will be definitely rebuilt in new test cases, so runner for it is postponed.
@mkalinin please, review

Issue #120, opened 2019-04-16
Some further SSZ work


Nashatyrev:
As the follow up to PR #115 I would like to do the following things: 
- add Incremental hashing of basic lists (packed)
- refactor `SSZBasicAccessors`
- make `ssz` module independent of others 
- do some performance benchs and probably pass some buffer for serialization to avoid creation of numerous `BytesValue`s objects 
- additional performance gain can be obtained by optimized serializing of `BytesValue` or `byte[]` SSZ fields

zilm13:
> make ssz module independent of others

@Nashatyrev 
It currently depends on `util` and `types`. What's wrong with such kind of dependency?

PR #122, opened 2019-04-23
Implement incremental packed list (basic type list/vector) hashing

Nashatyrev:
This is the completion of PR #115 
Also addresses one item of the issue #120 

No special tricks comparing to non-packed lists except the support for serializing subLists were added

### DIFF
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZIncrementalHasher.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZIncrementalHasher.java
@@ -68,7 +68,7 @@ public class SSZIncrementalHasher extends SSZSimpleHasher {
       } else if (!tracker.elementsUpdated.isEmpty()){
         if (type.isList() && ((SSZListType) type).getElementType().isBasicType()) {
           tracker.merkleTree =
-              updatePackedTrie((SSZListType) type, rawValue, childVisitor, tracker.merkleTree, tracker.elementsUpdated);
+              updatePackedTrie((SSZListType) type, rawValue, tracker.merkleTree, tracker.elementsUpdated);
         } else {
           tracker.merkleTree =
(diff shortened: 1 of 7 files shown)


ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZIncrementalHasher.java
```diff
@@ -88,19 +89,55 @@ private MerkleTrie updateNonPackedTrie(
...
+        idx -> childVisitor.apply(idx, type.getChild(value, idx)).getFinalRoot(),
+        type.getChildrenCount(value),
+        merkleTree,
+        elementsUpdated);
+  }
+
+  private MerkleTrie updatePackedTrie(
+      SSZListType type,
+      Object value,
+      BiFunction<Integer, Object, MerkleTrie> childVisitor,
```
mkalinin: It appears that `childVisitor` is not used.

ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZIncrementalHasher.java
```diff
@@ -162,4 +190,20 @@ private MerkleTrie copyWithSize(MerkleTrie trie, int newChunksCount) {
       return new MerkleTrie(newNodes);
     }
   }
+
+  private BytesValue serializePackedChunk(SSZListType basicListType, Object listValue, int chunkIndex) {
+    int typeSize = basicListType.getElementType().getSize();
+    int valsPerChunk = bytesPerChunk / typeSize;
+    if (valsPerChunk * typeSize != bytesPerChunk) {
+      throw new UnsupportedOperationException("");
```
mkalinin: Let's add some message here.

mkalinin (approved):
LGTM! There is a couple of cosmetic fixes might worth to be done.

zilm13 (approved):
LGTM

PR #130, opened 2019-05-11
Update deposit contract according to v0.6.1

zilm13:
Some quotations to better understand this update:
```
The deposit contract does little validation, pushing most of the validator onboarding logic to the beacon chain. In particular, the proof of possession (a BLS12-381 signature) is not verified by the deposit contract.
```

```
The proof for each deposit must be constructed against the deposit root contained in state.latest_eth1_data rather than the deposit root at the time the deposit was initially logged from the 1.0 chain. This entails storing a full deposit merkle tree locally and computing updated proofs against the latest_eth1_data.deposit_root as needed. See minimal_merkle.py for a sample implementation.
```
As you see this PR doesn't look like a production version of deposit root/proofs supplier, so I will be glad to give any feedback and help to move it to something we could really use in production.


zilm13:
Also I've not found this goddamn **** https://github.com/harmony-dev/beacon-chain-java/pull/130/files#diff-9b854443634f1745881ecae0970951edR174 anywhere except deposit contract sources. Am I correct it's really required and it couldn't be simplified?

Also some docs suggests using call of `get_deposit_root()` method from contract where it's possible but I'm not sure it'll simplify things a lot, especially when we are too far from current deposit event and everything is purged already in eth1.

pow/core/src/main/java/org/ethereum/beacon/pow/MinimalMerkle.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.stream.IntStream;
+
+import static tech.pegasys.artemis.util.bytes.BytesValue.concat;
+
+/**
+ * Minimal merkle tree <a
+ * href="https://en.wikipedia.org/wiki/Merkle_tree">https://en.wikipedia.org/wiki/Merkle_tree</a>
+ * implementation
+ */
+public class MinimalMerkle {
```
mkalinin: It looks like utility class for Merkle proof calculation. I remember your intention to swap it with an implementation that is used by SSZ hasher. It would be nice to have `MerkleTree` interface either.

pow/core/src/main/java/org/ethereum/beacon/pow/MinimalMerkle.java
```diff
@@ -0,0 +1,103 @@
...
+        zeroHashes[0] = Hash32.ZERO;
+      } else {
+        Hash32 lowerZeroHash = getZeroHash(distanceFromBottom - 1);
+        zeroHashes[distanceFromBottom] = hashFunction.apply(concat(lowerZeroHash, lowerZeroHash));
+      }
+    }
+    return zeroHashes[distanceFromBottom];
+  }
+
+  // # Compute a Merkle root of a right-zerobyte-padded 2**32 sized tree
```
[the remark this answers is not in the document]
zilm13: Done.

pow/core/src/main/java/org/ethereum/beacon/pow/MinimalMerkle.java
zilm13: I've checked it and I don't think it's a good idea, as one in hasher holds the tree but this one provides only utility classes as there are a lot of slicing operations here. Plus proofs feature will be  only here. I could extract interface later, when incremental version is added and keep both,  but I don't think we could have one interface with `org.ethereum.beacon.ssz.visitor.MerkleTrie`

pow/core/src/main/java/org/ethereum/beacon/pow/MerkleTree.java
```diff
@@ -0,0 +1,41 @@
...
+   */
+  ReadVector<Integer, Hash32> getProof(int index, int size);
+
+  /**
+   * Deposit Root of merkle tree with all elements up to index
+   *
+   * @param index last element index
+   * @return deposit root
+   */
+  Hash32 getDepositRoot(UInt64 index);
```
mkalinin: Let's remove `deposit` word from the name.

pow/core/src/main/java/org/ethereum/beacon/pow/MerkleTree.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Inserts value in tree / storage
+   *
+   * @param value Element value
+   */
+  void insertValue(V value);
```
mkalinin: I'd name it `add`, as `insert` better suites for method that able to add element somewhere in between the list.

zilm13:
@mkalinin done

util/src/main/java/org/ethereum/beacon/util/ConsumerList.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public final class ConsumerList<E> extends LinkedList<E> {
```
Nashatyrev: You are extending `LinkedList` and using another `LinkedList` instance as `delegate`. Does it make sense here?

zilm13:
@Nashatyrev ready for re-review

PR #141, opened 2019-05-24
[WIP] ssz: modify SOS to use end offset

zilm13:
One test is failing (`shouldHandleListList`), so WIP until research is finished

### DIFF
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SosDeserializer.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SosDeserializer.java
@@ -9,11 +9,11 @@ import org.ethereum.beacon.ssz.type.SSZListType;
 import org.ethereum.beacon.ssz.type.SSZType;
 import org.ethereum.beacon.ssz.visitor.SosDeserializer.DecodeResult;
 import org.javatuples.Pair;
+import tech.pegasys.artemis.util.bytes.BytesValue;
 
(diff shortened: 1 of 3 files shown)


zilm13:
Looks like it will not work in some cases: let me explain and give example.
Test: https://github.com/harmony-dev/beacon-chain-java/blob/78844923cea91733dbb23deec0de1ca42ddfa07d/ssz/src/test/java/org/ethereum/beacon/ssz/SSZSerializerTest.java#L207-L224

What happens, and when it fails:
Serializing goes ok:  
`0x280000001800000024000000240000000a0000000c000000616162620a0000000c00000063636464`

But when you are trying to deserialize it:
First, we decode `endOffset` of the list of the lists: `28000000`, 40 in decimal. Cool, all the rest are lists, reading there offsets:
- `18000000`, 1st list offset is 24, good
- `24000000`, 2nd list offset is 36. good. As we have only 36 bytes left it looks like criteria that fixed part is over, but waiit
- `24000000`, 3rd list offset is 36. Why it's the same as 2nd list offset? because 3rd list is empty.

So in both cases, either with `startOffset` or with `endOffset` we need one more pointer, to the end of the whole data in 1st case, or to the end of the fixed part.

mkalinin:
> am I correct in my resolution?

Yes, you are.
It's been a PoC for improvement proposed in https://github.com/ethereum/eth2.0-specs/issues/1087. Closing as it's been proved that this proposal is not viable.

PR #143, opened 2019-05-27
Speed up tests

zilm13:
Following attempts were done and measured to speed up tests execution:

1. gradle default, starting point: `7m 58s`
2. gradle #threads parallel: `7m 50s`
3. gradle #cores parallel: `7m 45s`
4. gradle #cores parallel + sszTests #cores parallel: `5m 6s`
5. gradle #cores parallel + sszTests (#cores - 1) parallel : `5m 38s`
6. default gradle, sszTests #cores parallel: `5m 2s`
7. default gradle, sszTests (#threads - 1) parallel: `4m 54s`

Gradle parallel test execution is a bit tricky in skipping `wire` submodule and may be other modules in future, however it doesn't provide any sensible improvement in execution time, maybe because it runs only module by module in parallel (waiting for all tests in module are completed) or, maybe, by another reason. So, I decided to apply only sszTests parallel running with  number of threads = number of cores (6).
@mkalinin, what do you think?

### DIFF
--- a/test/src/test/java/org/ethereum/beacon/test/SszTests.java
+++ b/test/src/test/java/org/ethereum/beacon/test/SszTests.java
@@ -34,7 +34,7 @@ public class SszTests extends TestUtils {
           SszStaticRunner testRunner = new SszStaticRunner(input.getValue0(), input.getValue1());
           return testRunner.run();
         },
-        getRecommendedThreadCount() // run it in parallel, a lot of tests
+        true // run it in parallel, a lot of tests
         );
   }
 }
--- a/test/src/test/java/org/ethereum/beacon/test/SszTests.java
+++ b/test/src/test/java/org/ethereum/beacon/test/SszTests.java
@@ -33,6 +33,8 @@ public class SszTests extends TestUtils {
         input -> {
           SszStaticRunner testRunner = new SszStaticRunner(input.getValue0(), input.getValue1());
           return testRunner.run();
-        });
+        },
+        getRecommendedThreadCount() // run it in parallel, a lot of tests
+        );
   }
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/ethereum/beacon/test/SszTests.java
```diff
@@ -33,6 +33,8 @@ public void testSszStatic() {
         input -> {
           SszStaticRunner testRunner = new SszStaticRunner(input.getValue0(), input.getValue1());
           return testRunner.run();
-        });
+        },
+        getRecommendedThreadCount() // run it in parallel, a lot of tests
```
mkalinin: IMO, it would be better to call `TestUtils.getTestExecutor(boolean parallel)` at lower layer with given `parallel` parameter.

test/src/test/java/org/ethereum/beacon/test/TestUtils.java
```diff
@@ -267,19 +281,43 @@ static BeaconChainSpec loadSpecByName(String name, boolean isBlsVerified) {
...
+              failed.set(true);
+            }
+            awaitLatch.countDown();
+          };
+      executor.execute(task);
     }
-    assertFalse(failed);
+
+    try {
+      awaitLatch.await();
```
mkalinin: You may use a default implementation of `Schedulers` interface to get an executor that returns `CompletableFuture`. It would be nice to combine them in a meta future to avoid this semaphore.

mkalinin (changes requested):

mkalinin (approved):
LGTM! 👍

PR #145, opened 2019-06-05
BeaconStateImpl set... list implementations

Nashatyrev:
I would better leave the existing list in place, since those `set...` methods can be potentially called (though shouldn't) by a user and in case of `ObservableList` updates tracking can be broken.


types/src/main/java/tech/pegasys/artemis/util/collections/WriteVector.java
```diff
@@ -1,5 +1,6 @@
 package tech.pegasys.artemis.util.collections;
 
+import java.util.Collection;
```
mkalinin, line 3: Looks like an accidental modification.

mkalinin (approved):
LGTM! Found a tiny cosmetic thing that would be nice to get rid of.

PR #148, opened 2019-06-11
Feature/validator rpc

zilm13:
It's base version with just two endpoints. Check 2 commits:
- Netty-reactor version https://github.com/harmony-dev/beacon-chain-java/commit/bc2434738d300ee2651d5a1a7b2a00ac429ac3ee
- Light4j version https://github.com/harmony-dev/beacon-chain-java/commit/ebbf836bc6ce24a0803956c42b60ad427d75ff22

As for me, I wouldn't recommend Light4j as it has very specific DSL, it's based on scaffold generation, there are no dependency injection mechanism for handlers, small user base etc. And as a cherry on top - you do everything by hands like json header.


zilm13:
Validator RPC made, only server part, client part with client logic (requesting, signing and submitting signed data) is a separate issue, not done here.
What should be noticed: 
- I don't like `SyncStatus` handling in `SyncManager`, especially possible concurrency issues, but I don't want to route all through `synchronized` method. Any better choices?
- Conversion to and from `IndexedAttestation` in `Server` code. Is it legit?

@mkalinin @Nashatyrev please, review

chain/src/main/java/org/ethereum/beacon/chain/observer/PendingOperations.java
```diff
@@ -1,20 +1,52 @@
...
-import org.ethereum.beacon.core.types.BLSPubkey;
-import org.ethereum.beacon.core.types.SlotNumber;
+
+import java.util.Collections;
+import java.util.List;
 
 /** A pending state interface. */
 public interface PendingOperations {
 
+  static PendingOperations getEmpty() {
```
mkalinin: You could implement it as a static variable and add `default` implementation to each method that would return empty list.

core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
```diff
@@ -21,6 +22,13 @@ public Bitfield(BytesValue bytes) {
     super(bytes);
   }
 
+  public Bitfield(Integer size, List<Integer> bits) {
+    super(BytesValue.wrap(new byte[(size + BYTE_SIZE - 1) / BYTE_SIZE]));
+    for (Integer bit : bits) {
+      setBit(bit, true);
```
mkalinin: If each bit is set to `true` why `bits` argument is needed at all?

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainAttester.java
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Creates {@link AttestationDataAndCustodyBit} instance and signs off on it.
+   *
+   * @param state a state at a slot that validator is attesting in.
+   * @param data an attestation data instance.
+   * @param signer message signer.
+   * @param spec Beacon chain spec
+   * @return signature of attestation data and custody bit.
+   */
+  static BLSSignature getAggregateSignature(
```
mkalinin: Why so static? :) I mean, why this code should be in a static method declared in top level interface?

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainAttester.java
```diff
@@ -0,0 +1,78 @@
...
+   * logic Prepares attestation with signature stubbed with zero BLS Signature Later signer could
+   * easily sign it
+   *
+   * @param validatorIndex index of the validator.
+   * @param shard shard number.
+   * @param observableState a state that attestation is based on.
+   * @param slot attestation slot
+   * @return unsigned attestation
+   */
+  Attestation prepareAttestation(
```
mkalinin: I'd rather move this method to another interface and share its implementation between attester and other dependent components.

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainProposer.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Signs off on a block.
+   *
+   * @param state state at the slot of created block.
+   * @param block block
+   * @param signer message signer.
+   * @param spec Beacon chain spec
+   * @return signature of proposal signed data.
+   */
+  static BLSSignature getProposalSignature(
```
mkalinin: Same here. Maybe there is a time to break it down into several interfaces? This change makes significant sense for segregated validator instance.

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainProposer.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Prepares builder with complete block without signature to sign it off later. Part of {@link
+   * #propose(ObservableBeaconState, MessageSigner)} logic without signing and distribution
+   *
+   * @param slot Slot number we are going to create block for
+   * @param randaoReveal Signer RANDAO reveal
+   * @param observableState A state on top of which new block is created.
+   * @return builder with unsigned block, ready for sign and distribution
+   */
+  BeaconBlock.Builder prepareBuilder(
```
mkalinin: I think it's better to return a block rather than a builder. Further, builder can be easily instantiated with a block if this is desirable case.

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainProposer.java
```diff
@@ -0,0 +1,90 @@
...
+      SlotNumber slot, BLSSignature randaoReveal, ObservableBeaconState observableState);
+
+  /**
+   * Given a state returns graffiti value.
+   *
+   * @param state a state.
+   * @return graffiti value.
+   */
+  default Bytes32 getGraffiti(BeaconState state) {
```
mkalinin: Is it used anywhere? Or assume to be used later on?

mkalinin (changes requested):

core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
zilm13: we occupy size bits and mark only those that are in list to true

chain/src/main/java/org/ethereum/beacon/chain/observer/PendingOperations.java
zilm13: Nice advice!

validator/embedded/src/main/java/org/ethereum/beacon/validator/BeaconChainProposer.java
zilm13: Looks like a hack made to point out that it's fake

zilm13:
@mkalinin please, review

core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
```diff
@@ -10,6 +10,7 @@
 
 @SSZSerializable(serializeAs = BytesValue.class)
 public class Bitfield extends DelegatingBytesValue {
+  private final static int BYTE_SIZE = Byte.SIZE;
```
mkalinin: Do we need this extra constant? Why not just use `Byte.SIZE` in the code?

validator/server/src/main/java/org/ethereum/beacon/validator/api/InvalidInputException.java
```diff
@@ -0,0 +1,17 @@
+package org.ethereum.beacon.validator.api;
+
+public class InvalidInputException extends RuntimeException {
```
mkalinin, line 4: It would be nice to have a description for these exception.

validator/server/src/main/java/org/ethereum/beacon/validator/api/RestClient.java
```diff
@@ -0,0 +1,112 @@
...
+import org.ethereum.beacon.validator.api.model.ValidatorDutiesResponse;
+
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.math.BigInteger;
+
+public class RestClient {
```
mkalinin: This class looks like an implementation of lacking `ValidatorClient` interface. Would you mind to add this interface too?

validator/server/src/main/resources/rest-server.properties
```diff
@@ -0,0 +1 @@
+versionNumber='0.0.2'
```
mkalinin: I think we can use `ClientInfo` as a source of version, can't we?

mkalinin: It would only need to be abstracted from `start` module.

wire/src/main/java/org/ethereum/beacon/wire/SimplePeerManagerImpl.java
```diff
@@ -86,6 +93,15 @@ protected HelloMessage createLocalHello() {
         head.getBlock().getSlot());
   }
 
+  private void updateBestSlot(PeerImpl peer) {
```
mkalinin, line 96: I think it would be more preferable if this function was mapped onto `connectedPeerStream` and store best slot in some dedicated component.

wire/src/main/java/org/ethereum/beacon/wire/sync/SyncManager.java
```diff
@@ -16,4 +17,53 @@
   void setSyncApi(WireApiSync syncApi);
 
   Publisher<Feedback<BeaconBlock>> getBlocksReadyToImport();
+
+  void start();
+
+  void stop();
+
+  Publisher<SyncMode> getSyncModeStream();
```
mkalinin, line 25: Could this stream be superseded with status stream?

wire/src/main/java/org/ethereum/beacon/wire/sync/SyncManager.java
```diff
@@ -16,4 +17,53 @@
...
+
+  Publisher<SyncStatus> getSyncStatusStream();
```
mkalinin: I think it would be more reactive wise if we have several outbound streams that exposes information with lower granularity, like best imported slot, sync mode, etc. And then another component that collects this data and merges them into a stream of sync status.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ReactorNettyServer.java
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Parses params from request uri and returns key value map. Several values per key supported
+   *
+   * @param request Server request
+   * @return GET query parameters ?key1=value1&key2=value2&key2=value3 -> key1: [value1], key2:
+   *     [value2, value3]
+   * @throws URISyntaxException when couldn't parse uri
+   * @throws UnsupportedEncodingException when couldn't parse uri
+   */
+  static Map<String, List<String>> getRequestParams(HttpServerRequest request)
```
mkalinin: These util methods should be out of this class, sited in a separated utility class.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ReactorNettyServer.java
```diff
@@ -0,0 +1,183 @@
...
+      assert params.containsKey(param);
+      return params.get(param);
+    } catch (AssertionError e) {
+      throw new InvalidInputException(
+          String.format("Parameters map doesn't contain required param `%s`", param));
+    }
+  }
+
+  /** Adopted from https://stackoverflow.com/a/13592567 */
+  private static Map<String, List<String>> splitQuery(URI uri) throws UnsupportedEncodingException {
```
mkalinin: I am not much in favor of this. We have to maintain HTTP related functionality implemented by many other libraries that have already counted all pitfalls and solved problems that could potentially arise here and there.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ReactorNettyServer.java
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * POST Endpoint helper
+   *
+   * <p>Executes func with `HttpServerRequest` input and returns 200 OK if everything was ok
+   *
+   * <p>Responses with 202 ACCEPTED when catches {@link PartiallyFailedException} from func function
+   *
+   * <p>Responses with 400 BAD REQUEST when catches {@link InvalidInputException} from func function
+   */
+  static Publisher<Void> processPostRequest(
```
mkalinin: IMO, these functionality should be a part of pipeline driven by reactive streams. Exception handler should, also, be a part of that pipeline.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ReactorNettyServer.java
```diff
@@ -0,0 +1,183 @@
...
+
+  public ReactorNettyServer(String serverHost, Integer serverPort) {
+    serverRes = LoopResources.create("server");
+    HttpServer httpServer =
+        HttpServer.create()
+            .tcpConfiguration(tcpServer -> tcpServer.runOn(serverRes))
+            .host(serverHost)
+            .port(serverPort);
+    final HttpServer serverWithRoutes = addRoutes(httpServer);
+    this.server = serverWithRoutes.bindNow();
```
mkalinin: I think this is a common practice to start something outside of constructor via `start` method. Would you mind to follow it in this case?

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
```diff
@@ -0,0 +1,389 @@
...
+                time = observableBeaconState.getLatestSlotState().getGenesisTime();
+              }
+            });
+
+    Flux.from(syncManager.getSyncStatusStream())
+        .subscribe(
+            syncStatus -> {
+              shortSync = SyncManager.SyncMode.Short.equals(syncStatus.getSyncMode());
+              syncingResponse =
+                  new SyncingResponse(
```
mkalinin: We might want to have a `create` method that creates `SyncingResponse` from `SyncStatus` object. It would help to keep code clean.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
```diff
@@ -0,0 +1,389 @@
...
+                          ? BigInteger.ZERO
+                          : syncStatus.getCurrent().toBI(),
+                      syncStatus.getBestKnown() == null
+                          ? BigInteger.ZERO
+                          : syncStatus.getBestKnown().toBI());
+            });
+  }
+
+  @Override
+  HttpServer addRoutes(HttpServer httpServer) {
```
mkalinin: As for me router is a separate instance that holds a map of routes onto controller methods. Something like `route("/node/version", versionController::index)`. Each of controller's method supports same interface, e.g. `public Response index(Request request)`.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
```diff
@@ -0,0 +1,389 @@
...
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private String produceForkResponse() {
+    try {
+      ForkResponse forkResponse =
+          new ForkResponse(
+              observableBeaconState.getLatestSlotState().getFork().getCurrentVersion().toString(),
```
mkalinin: There could be a race condition here in a case when `observableBeaconState` is updated somewhere in between of these subsequent calls.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
```diff
@@ -0,0 +1,389 @@
...
+
+      return props;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private String produceSyncingResponse() {
+    try {
+      return mapper.writeValueAsString(syncingResponse);
```
mkalinin: It would be handy if this mapping was segregated and put into a middleware pipeline.

core/src/main/java/org/ethereum/beacon/core/types/Bitfield.java
zilm13: removed class

wire/src/main/java/org/ethereum/beacon/wire/SimplePeerManagerImpl.java
zilm13: I don't understand, who else should care about best slot of all peers? Manager. I've chose `SimplePeerManagerImpl` because we are not required to start sync to know it. Separate `MaxKnowBestSlotManager`, really? It looks like over-engineering.

mkalinin: This is single responsibility. I think there should be a component that accumulates knowledge about the chain we're currently connected to. It could be not only max known slot. You may leave it as it is for now. It would be just cleaner if peers management and chain attributes were separated.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
zilm13: I see such simple resolution for it: https://github.com/harmony-dev/beacon-chain-java/commit/74342ba7b0735bb58a6da3e9f7b9bb7b0e88e316
but I don't like it much, because it's running on one thread which is used for server answers and shouldn't be blocked. @Nashatyrev do you have any idea for such cases?

wire/src/main/java/org/ethereum/beacon/wire/PeerManager.java
```diff
@@ -1,48 +1,46 @@
...
-  /**
-   * Stream of new peer connections
-   */
+  /** Stream of new peer connections */
   Publisher<Peer> connectedPeerStream();
 
   /**
-   * Stream of peer disconnects
-   * Peer must occur in this stream strictly after occurring in the {@link #connectedPeerStream()}
+   * Stream of peer disconnects Peer must occur in this stream strictly after occurring in the
```
mkalinin: Let's add `<p>` tag before `Peer...`

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
```diff
@@ -0,0 +1,107 @@
...
+            new AttestationSubmitController(
+                syncManager, peerManager.getWireApiSub(), stateProcessor, spec)));
+    controllers.add(
+        ControllerRoute.of(GET, "/node/fork", new ForkController(stateProcessor, spec, chainId)));
+    this.server = new RestServerVerticle(serverPort, controllers);
+  }
+
+  @Override
+  public synchronized void start() {
+    logger.info("Starting Validator REST on {} port", server.getServerPort());
```
mkalinin, line 83: We might want to have double start protection.

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/RestController.java
```diff
@@ -0,0 +1,75 @@
...
+import io.vertx.core.MultiMap;
+import io.vertx.core.json.Json;
+import io.vertx.ext.web.RoutingContext;
+import org.ethereum.beacon.validator.api.InvalidInputException;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+public abstract class RestController implements Controller {
+  static String getParamString(String param, MultiMap params) throws InvalidInputException {
```
mkalinin: I think it would be better to move these methods to a kind of utility class.

validator/server/src/main/java/org/ethereum/beacon/validator/api/ValidatorRest.java
Nashatyrev: In this concrete case (no blocking calls, no calls to external components) both solutions are fine:
- synchronize both methods (as you already did in `ForkController`)
- just assign `observableBeaconState` to a local variable

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/RestController.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  static List<String> getParamStringList(String param, MultiMap params)
+      throws InvalidInputException {
+    try {
+      assert params.contains(param);
+      return params.getAll(param);
+    } catch (AssertionError e) {
+      throw new InvalidInputException(
```
mkalinin: What do you think about `parseListParam`, `parseParam` with corresponding validation as a part of these methods execution?

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/RestController.java
```diff
@@ -0,0 +1,75 @@
...
+          String.format("Parameters map doesn't contain 1 required param `%s`", param));
+    }
+  }
+
+  static List<String> getParamStringList(String param, MultiMap params)
+      throws InvalidInputException {
+    try {
+      assert params.contains(param);
```
mkalinin: `if (!params.contains(param)) throw` should be cheaper than throwing assertion error.

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/SyncRestController.java
```diff
@@ -0,0 +1,98 @@
...
+   * <p>Responses with 400 BAD REQUEST when catches {@link InvalidInputException} from func function
+   *
+   * <p>Responses with 406 NOT ACCEPTABLE when catches {@link NotAcceptableInputException} from func
+   * function
+   *
+   * <p>Responses with 500 INTERNAL SERVER ERROR for all other errors
+   */
+  Handler<RoutingContext> processGetRequestImpl(Function<HttpServerRequest, Object> response) {
+    return event -> {
+      if (!shortSync) {
```
mkalinin, line 38: This logic makes this general purpose controller too specific. What if we need errors processing and don't need a check for sync status? I think it would be better to move this check out of this controller's scope.

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/BlockSubmitController.java
```diff
@@ -0,0 +1,55 @@
...
+import org.ethereum.beacon.core.BeaconBlock;
+import org.ethereum.beacon.validator.api.InvalidInputException;
+import org.ethereum.beacon.validator.api.PartiallyFailedException;
+import org.ethereum.beacon.validator.api.model.BlockSubmit;
+import org.ethereum.beacon.wire.WireApiSub;
+import org.ethereum.beacon.wire.sync.SyncManager;
+
+import java.util.Optional;
+
+public class BlockSubmitController extends SyncRestController {
```
mkalinin: We might want to merge this controller with `BlockController` and keep following convention with single controller for all interactions with same entity.

mkalinin (changes requested):
Looks good! 👍  I think a few more improvements should make it more extensible for upcoming changes.

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/SyncRestController.java
zilm13: If we need something else, we just extend its parent, `RestController.java` and use it. It's just a page of code, we couldn't make it universal for all cases we cannot foresee.

validator/server/src/main/java/org/ethereum/beacon/validator/api/controller/DutiesController.java
```diff
@@ -0,0 +1,106 @@
...
+      SyncManager syncManager,
+      BeaconChainSpec spec) {
+    super(syncManager);
+    Flux.from(stateProcessor.getObservableStateStream()).subscribe(this::updateState);
+    this.spec = spec;
+    this.service = service;
+  }
+
+  private synchronized void updateState(ObservableBeaconState observableBeaconState) {
+    this.observableBeaconState = observableBeaconState;
```
mkalinin, line 50: I think in this case it would be more efficient if a local copy of `observableBeaconState` were used to read the data from, `synchronized` might be an overkill here.

mkalinin (approved):

PR #149, opened 2019-06-21
Update consensus to v0.7.0

mkalinin:
### What was done?
Updates consensus to v0.7.1. Among other things are:
- Deposit logic update related to https://github.com/ethereum/eth2.0-specs/pull/1152

_Note:_ moved `ChainStart` class to `consensus` module to get rid of circular dependency. Not a perfect solution. We may want to re-design our modules later on.


mkalinin:
**Note:** part of state and ssz tests are ignored and requires and update.

Nashatyrev (approved):

core/src/main/java/org/ethereum/beacon/core/operations/attestation/Crosslink.java
```diff
@@ -12,68 +12,87 @@
...
-  public static final Crosslink EMPTY = new Crosslink(EpochNumber.ZERO, Hash32.ZERO, Hash32.ZERO);
+  public static final Crosslink EMPTY =
+      new Crosslink(ShardNumber.ZERO, EpochNumber.ZERO, EpochNumber.ZERO, Hash32.ZERO, Hash32.ZERO);
 
-  /** Epoch number. */
-  @SSZ private final EpochNumber epoch;
+  /** Shard number. */
+  @SSZ private final ShardNumber shard;
+  /** Crosslinking data from epochs [start....end-1]. */
+   @SSZ private final EpochNumber startEpoch;
```
zilm13: Space!
```suggestion
@SSZ private final EpochNumber startEpoch;
```

validator/src/main/java/org/ethereum/beacon/validator/proposer/BeaconChainProposerImpl.java
```diff
@@ -155,6 +155,7 @@ private Eth1Data getEth1Data(BeaconState state) {
         .filter(
             eth1Data ->
                 depositContract.hasDepositRoot(eth1Data.getBlockHash(), eth1Data.getDepositRoot()))
+        // TODO throw exception if contract data can't be red
```
zilm13:
```suggestion
        // TODO throw exception if contract data can't be read
```

zilm13 (changes requested):
I'm not sure you are able to deal with it

PR #151, opened 2019-06-24
Feature/validator rpc 1

Nashatyrev:
IMHO this is more canonical RX stream creation. It should be also safer to use stateless approach
@zilm13 Can you please test this if looks ok?


zilm13 (approved):
LGTM

Issue #146, opened 2019-06-07
Github test debts


zilm13:
Current debts, both blocked by 3rd party issue resolution:

- [x] bls, one case (blocked by https://github.com/ethereum/eth2.0-spec-tests/issues/3 )
- [x] spec configs in tests repository (blocked by https://github.com/ethereum/eth2.0-spec-tests/issues/1)

Please, follow updates and resolve when possible

PR #152, opened 2019-06-24
Update github tests to v0.7.1

zilm13:
Everything is ok, plus one item from https://github.com/harmony-dev/beacon-chain-java/issues/146 is fixed


mkalinin:
we're passing all tests! 🎉 
that is sooo cool!

PR #153, opened 2019-06-26
Add SSZ Union type support 

Nashatyrev:
Please read `ReadUnion` class javadoc for usage examples

### DIFF
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
@@ -254,23 +254,15 @@ public class SSZBuilder {
 
   public SSZBuilder addDefaultContainerAccessors() {
     checkAlreadyInitialized();
-    containerAccessors.add(this::createDefaultContainerAccessor);
+    containerAccessors.add(() -> new SimpleContainerAccessor(sszSchemeBuilder, objCreator));
     return this;
-  }
-
-  private SSZContainerAccessor createDefaultContainerAccessor() {
-    return new SimpleContainerAccessor(sszSchemeBuilder, objCreator);
   }
 
   public SSZBuilder addDefaultUnionAccessors() {
     checkAlreadyInitialized();
-    unionAccessors.add(GenericTypeSSZUnionAccessor::new);
-    unionAccessors.add(this::createDefaultUnionAccessor);
+    unionAccessors.add(() -> new GenericTypeSSZUnionAccessor());
+    unionAccessors.add(() -> new SchemeSSZUnionAccessor(sszSchemeBuilder));
     return this;
-  }
-
-  private SSZUnionAccessor createDefaultUnionAccessor() {
-    return new SchemeSSZUnionAccessor(sszSchemeBuilder);
   }
 
   /**
--- a/ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
+++ b/ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
@@ -27,8 +27,8 @@ import org.ethereum.beacon.ssz.access.list.ArrayAccessor;
 import org.ethereum.beacon.ssz.access.list.BytesValueAccessor;
 import org.ethereum.beacon.ssz.access.list.ListAccessor;
 import org.ethereum.beacon.ssz.access.list.ReadListAccessor;
-import org.ethereum.beacon.ssz.access.union.GenericTypeUnionAccessor;
-import org.ethereum.beacon.ssz.access.union.UnionAccessor;
+import org.ethereum.beacon.ssz.access.union.GenericTypeSSZUnionAccessor;
+import org.ethereum.beacon.ssz.access.union.SchemeSSZUnionAccessor;
 import org.ethereum.beacon.ssz.annotation.SSZ;
 import org.ethereum.beacon.ssz.annotation.SSZSerializable;
 import org.ethereum.beacon.ssz.annotation.SSZTransient;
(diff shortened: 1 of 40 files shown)


ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZUnionAccessor.java
```diff
@@ -0,0 +1,27 @@
+package org.ethereum.beacon.ssz.access;
+
+import java.util.List;
+
+/**
+ * Handles containers (aka structures or ordered heterogenous collection of values),
```
zilm13: JavaDoc is a copy of `SSZContainerAccessor`

ssz/src/main/java/org/ethereum/beacon/ssz/SSZSerializer.java
```diff
@@ -61,13 +59,10 @@ public SerializerResult visitList(SSZListType descriptor, Object listValue, int
...
    */
   public <C> C decode(byte[] data, Class<? extends C> clazz) {
-    DecodeResult decodeResult = sszVisitorHost.handleAny(
+    Object decodeResult = sszVisitorHost.handleAny(
         typeResolver.resolveSSZType(new SSZField(clazz)),
-        Pair.with(Bytes.wrap(data), null),
+        Bytes.wrap(data),
         new SosDeserializer());
-    if (data.length != decodeResult.readBytes) {
-      throw new SSZSerializeException(String.format("Invalid SSZ encoding, calculated data size %s bytes, while provided %s bytes", decodeResult.readBytes, data.length));
```
zilm13, line 69: Do I understand correctly that with addition of union to ssz, we couldn't keep this check at all?

zilm13 (approved):
LGTM

ssz/src/main/java/org/ethereum/beacon/ssz/SSZSerializer.java
Nashatyrev: This is about sos not about unions. When decoding, the end of the latest variable part is determined from the byte array length. So if we have erroneous extra bytes, then they either would be decoded to erroneous extra list elements (and this can't be detected) or the fail would happen during decoding of fixed size value

ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
```diff
@@ -254,6 +262,17 @@ private SSZContainerAccessor createDefaultContainerAccessor() {
     return new SimpleContainerAccessor(sszSchemeBuilder, objCreator);
   }
 
+  public SSZBuilder addDefaultUnionAccessors() {
```
mkalinin, line 261: Naming looks confusing a bit. Too many `default` words used in one place. For instance, does defaultness degree of `GenericTypeSSZUnionAccessor` differ from accessor created with `createDefaultUnionAccessor`?

ssz/src/main/java/org/ethereum/beacon/ssz/access/union/GenericTypeSSZUnionAccessor.java
```diff
@@ -0,0 +1,78 @@
...
+    };
+  }
+
+  @Override
+  public UnionAccessor getAccessor(SSZField compositeDescriptor) {
+    return new GenericUnionAccessor(compositeDescriptor);
+  }
+
+  private List<SSZField> getChildDescriptorsFromGenericType(SSZField unionDescriptor) {
+    if (!ReadUnion.GenericTypedUnion.class.isAssignableFrom(unionDescriptor.getRawClass())) {
```
mkalinin: This check is called from at least two places. I think `isSupported(Class clazz)` helper method is able to make it more readable.

ssz/src/main/java/org/ethereum/beacon/ssz/SSZSerializer.java
zilm13: Got it

types/src/main/java/tech/pegasys/artemis/util/collections/ReadUnion.java
```diff
@@ -0,0 +1,128 @@
...
+ * It may hold one of predefined type value or special Null value
+ * It also holds the index of type currently stored in this structure
+ *
+ * For SSZ serialization/deserialiazation/hashing this type can be used in
+ * one of the following means:
+ *
+ * Direct subclassing with declaring all the SSZ Union types explicitly like the following:
+ *
+ * <code>
+      @SSZSerializable
```
mkalinin: This part without asterisks is not a valid java doc.

types/src/main/java/tech/pegasys/artemis/util/collections/ReadUnion.java
```diff
@@ -0,0 +1,128 @@
...
+      }
+ * </code>
+ *
+ * Please note that the latter usecase is only possible when the Union is a member of another SSZ
+ * container since only in this case generic types specialization is available at runtime
+ *
+ * This class denotes read-only Union. See {@link WriteUnion} for read-write version and
+ * {@link UnionImpl} for implementing class
+ */
+public interface ReadUnion {
```
mkalinin: `ReadUnion` reads a bit odd. Maybe use `Union/MutableUnion` notation?

mkalinin: Union type is not a collection by its semantics. IMO, better to move it out of `collection` package to e.g. `union`.

mkalinin (approved):
LGTM! Just have several notes regarding naming.

types/src/main/java/tech/pegasys/artemis/util/collections/ReadUnion.java
Nashatyrev: Ok, renamed. Just named them the same way as `ReadList/WriteList`. `ReadList` was named to avoid confusion with `java.util.List`
Let's leave them in the `collection` package. Union is still a degenerate case of collection :))

ssz/src/main/java/org/ethereum/beacon/ssz/SSZBuilder.java
Nashatyrev: Yep, total mess :) Fixed.

PR #154, opened 2019-06-27
Segregated signer

mkalinin:
### What was done?
A bit different approach in segregating signer from proposer and attester.
Inroduces following abstractions:
- block signer
- randao generator
- attestation signer

Proposer and attester become responsible for creating blocks and attestations with empty signature. Signers take care of signature and randao generator produces new reveals. IMO, such design should fit the needs of separated validator instance.


zilm13 (approved):
LGTM

Issue #63, opened 2019-02-07
Assert local time is correct


mkalinin:
### To Do
Beacon chain processing is a time critical process. Hence, local clock time should be verified against some source on the Internet. User should be notified if local clock time differs from the global one.

Nashatyrev:
Apache NTP client: https://commons.apache.org/proper/commons-net/

zilm13:
I think app should have it's own time or maybe choice in config between own and system.
Plus time correction is not start only task. If your client is running for a days, it's definitely a regular job.

PR #156, opened 2019-07-02
[WIP] Features safe time

zilm13:
It's a draft to solve the #63 issue.
The idea is to have several time streams: `SystemTime` with system time, `NetworkTime` with NTP servers corrected system time, `StatisticsTime` with time estimated from network objects such as blocks.
And next we have `TimeStrategy`'s designed to provide safe `Time` ticker for slot production and other events. Normally we use `NetworkTime` for it but if the difference between network time and statistics time is too big we suppose that network time is under attack, so we are starting to use `StatisticsTime` as node time source. We are not using `StatisticsTime` normally because it has a lot of variance and bad precision for usual days.


util/time/src/main/java/org/ethereum/beacon/time/provider/SystemTime.java
```diff
@@ -0,0 +1,49 @@
...
+public class SystemTime implements TimeProvider {
+  private static final int MILLIS_IN_SEC = 1000;
+  private final SimpleProcessor<Time> timeProcessor;
+  private final AtomicLong latest = new AtomicLong(-1);
+
+  public SystemTime(Scheduler events, Scheduler worker) {
+    this.timeProcessor = new SimpleProcessor<Time>(events, "TimeProvider.system");
+    worker.executeR(
+        () -> {
+          while (!Thread.interrupted()) {
```
mkalinin:
```suggestion
          while (!Thread.currentThread().isInterrupted()) {
```

util/time/src/main/java/org/ethereum/beacon/time/provider/SystemTime.java
```diff
@@ -0,0 +1,49 @@
...
+    if (currentSec > latest.get()) {
+      System.out.println("Time: " + currentSec);
+      timeProcessor.onNext(Time.of(currentSec));
+      latest.set(currentSec);
+    }
+    long currentMs = Instant.now().getLong(ChronoField.MILLI_OF_SECOND);
+    try {
+      Thread.sleep(MILLIS_IN_SEC - currentMs);
+    } catch (InterruptedException e) {
+      throw new RuntimeException("SystemTime stream thread interrupted!", e);
```
mkalinin:
```suggestion
      Thread.currentThread().interrupt();
```

In this case  interrupted thread state will be escalated and handled on the higher level as you already did it in a `while` loop.

util/time/src/main/java/org/ethereum/beacon/time/NetworkFirstStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+  private final AtomicLong latestStatistics = new AtomicLong(-1);
+  private static final Logger logger = LogManager.getLogger("time");
+
+  public NetworkFirstStrategy(
+      Scheduler scheduler,
+      NetworkTime networkTime,
+      StatisticsTime statisticsTime,
+      int allowedDelta) {
+    this.timeProcessor = new SimpleProcessor<Time>(scheduler, "NetworkFirstStrategy");
+    Flux.from(networkTime.getTimeStream())
```
mkalinin, line 31: May we have a component with two inbound streams and a single processor that handles them both?

util/time/src/main/java/org/ethereum/beacon/time/provider/NetworkTime.java
```diff
@@ -0,0 +1,134 @@
...
+    long currentSec = nowCorrected.getEpochSecond();
+    long currentMs = nowCorrected.getLong(ChronoField.MILLI_OF_SECOND);
+    if (currentSec > latest.get()) {
+      timeProcessor.onNext(Time.of(currentSec));
+      latest.set(currentSec);
+    }
+    try {
+      Thread.sleep(MILLIS_IN_SEC - currentMs);
+    } catch (InterruptedException e) {
+      throw new RuntimeException("NetworkTime stream thread interrupted!", e);
```
mkalinin:
```suggestion
      Thread.currentThread().interrupt();
```

util/time/src/main/java/org/ethereum/beacon/time/provider/NetworkTime.java
```diff
@@ -0,0 +1,134 @@
...
+      Scheduler events,
+      Scheduler worker,
+      Scheduler networkWorker,
+      List<String> ntpServers,
+      long correctionPeriodMs) {
+    this.ntpServers = ntpServers;
+    this.timeProcessor = new SimpleProcessor<Time>(events, "TimeProvider.network");
+    worker.executeR(
+        () -> {
+          while (!Thread.interrupted()) {
```
mkalinin:
```suggestion
          while (!Thread.currentThread().isInterrupted()) {
```

mkalinin (commented):
Quite a good start in general. I left some comments and suggestions that could be useful.

PR #158, opened 2019-07-09
Bitlist, Bitvector, max list size

zilm13:
At first max list size will be added using ssz annotation with expected cons (max size is checked only during serialization).
Current state:
- [ ] Observation helper fixes to go
- [x] Test fixes to go


wire/src/main/java/org/ethereum/beacon/wire/sync/SyncQueueImpl.java
```diff
@@ -49,12 +50,14 @@ public SyncQueueImpl(BeaconBlockTree blockTree) {
             slot = finalBlock.getSlot();
           }
           sink.next(new BlockRequest(slot, null, maxBlocksRequest, false, 0));
-          return slot.plus(maxBlocksRequest);
+          SlotNumber tmp =  slot.plus(maxBlocksRequest);
+          System.out.println(String.format("Request: %s, %s, %s, %s, %s", tmp, slot, finalBlock, maxHeightFromFinal, maxBlocksRequest));
+          return tmp;
```
Nashatyrev: Debug out left

core/src/main/java/org/ethereum/beacon/core/types/Bitvector.java
```diff
@@ -0,0 +1,176 @@
...
+import tech.pegasys.artemis.util.bytes.BytesValue;
+import tech.pegasys.artemis.util.bytes.DelegatingBytesValue;
+import tech.pegasys.artemis.util.bytes.MutableBytesValue;
+import tech.pegasys.artemis.util.uint.UInt64;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@SSZSerializable(listAccessor = Bitvector.BitVectorAccessor.class)
+public class Bitvector extends DelegatingBytesValue {
```
Nashatyrev: Bitlist & Bitvector looks pretty duplicating each other. M.b. makes sense to extract a common subclass?

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
```diff
@@ -0,0 +1,221 @@
...
+import tech.pegasys.artemis.util.bytes.BytesValue;
+import tech.pegasys.artemis.util.bytes.BytesValues;
+import tech.pegasys.artemis.util.bytes.DelegatingBytesValue;
+import tech.pegasys.artemis.util.bytes.MutableBytesValue;
+import tech.pegasys.artemis.util.uint.UInt64;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@SSZSerializable(listAccessor = Bitlist.BitListAccessor.class)
```
Nashatyrev: I would name classes `BitList` & `BitVector`. This would look more like Java names

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
```diff
@@ -0,0 +1,221 @@
+package org.ethereum.beacon.core.types;
```
Nashatyrev: I would love to see `Bitlist` & `Bitvector` classes in the `types` module, since the structures are not Eth specific

core/src/main/java/org/ethereum/beacon/core/spec/SpecConstantsResolver.java
```diff
@@ -23,12 +23,18 @@ public SpecConstantsResolver(SpecConstants constants) {
...
           && method.getName().startsWith("get")
-          && Number.class.isAssignableFrom(method.getReturnType())) {
+          && isNumber(method.getReturnType())) {
         constMethods.put(method.getName(), method);
       }
     }
   }
 
+  private boolean isNumber(Class type) {
+    return type == int.class
```
mkalinin, line 34: I'd add `byte` type is well, just in case.

Nashatyrev (changes requested):

ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZListAccessor.java
```diff
@@ -1,44 +1,70 @@
...
-    /**
-     * Appends a new child to the building list
-     */
-    void addChild(Object childValue);
+  /**
+   * Returns atomic size of a list. Normally it's 1:1 to element size, see {@link
+   * #getChildrenCount(Object)}, but assuming structures like Bit List we have elements that are
+   * smaller than byte (atoms) which are accessible as bytes
+   */
+  default int getAtomicChildrenCount(Object value) {
```
mkalinin: This change breaks an invariant of `getChildrenCount` that should return a number of bits for bit list and vector like it does for other collections. I think we need a calculator for size in bytes for each type which defaults to `sizeof(element) * getChildrenCount(value)`. And, also, we might need a stream that returns encoded bytes for collection.

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
zilm13: I did it at first, but later find out that in spec they are `Bitlist` and `Bitvector`, so I decided to have naming like in spec

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
zilm13: I did it at first, but later find out that in spec they are `Bitlist` and `Bitvector`, so I decided to have naming like in spec

ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZListAccessor.java
```diff
@@ -1,44 +1,70 @@
...
 
   @Override
-  ListInstanceBuilder createInstanceBuilder(SSZField listType);
+  ListInstanceBuilder createInstanceBuilder(SSZType sszType);
+
+  /**
+   * Some types of lists have extra size information encoded during serialization, which should be
+   * removed for hashing. Just use this method.
+   */
+  default BytesValue removeListSize(Object value, BytesValue serialization) {
```
mkalinin: We might want to generalize this. `removeListSize` is something too specific for interface of such a high level.

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
zilm13: We will need to move accessor to `ssz` module in this case or we are getting cyclic dependency

core/src/main/java/org/ethereum/beacon/core/types/Bitvector.java
zilm13: I will try

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    int neededBytes = (size + 7) / 8;
+    return new Bitlist(size, blank.toBytesValue().slice(0, neededBytes), maxSize);
+  }
+
+  private static BytesValue checkSize(BytesValue input, int size) {
+    int neededBytes = (size + 7) / 8;
+    if (neededBytes != input.size()) {
+      throw new IllegalArgumentException(
```
mkalinin: I think this exception could be reflected in SSZ code as `SSZDeserializeException` for conformance.

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
```diff
@@ -0,0 +1,221 @@
...
+  public String toString() {
+    StringBuilder ret = new StringBuilder("0b");
+    for (int i = 0; i < byteSize(); i++) {
+      ret.append(toBinaryString(get(i)));
+    }
+    return ret.toString();
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
mkalinin: `equals` and `hashCode` seems to be redundant here as they are already defined in `DelegatingBytesValue`.

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
```diff
@@ -0,0 +1,221 @@
...
+    // bits to bytes
+    return (size + 7) / 8;
+  }
+
+  private int fullByteSize() {
+    // bits to bytes + size bit
+    return (size + 8) / 8;
+  }
+
+  public static class BitListAccessor extends AbstractListAccessor {
```
mkalinin: It should be moved to `ssz` module.

mkalinin (changes requested):

ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZListAccessor.java
Nashatyrev: I would suggest to change the `SSZType.getSize()` to return rational value (or `double` for simplicity, but you need to be careful with converting to int) and add a `SSZBitType` as the `Bitlist` element type with the size of `1/8`

ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZListAccessor.java
Nashatyrev: Yep, Accessors has nothing with serialization :(

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
zilm13: `DelegatingBytesValue` doesn't contain actual size and maxSize

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
zilm13: Ok

ssz/src/main/java/org/ethereum/beacon/ssz/access/SSZListAccessor.java
zilm13: An extra complexity that this calculation should be different for serialization and hasher

core/src/main/java/org/ethereum/beacon/core/types/Bitlist.java
mkalinin: Got it!

zilm13:
@mkalinin @Nashatyrev let's do it again!

ssz/src/main/java/org/ethereum/beacon/ssz/access/list/BitlistAccessor.java
```diff
@@ -0,0 +1,61 @@
...
+import tech.pegasys.artemis.util.bytes.MutableBytesValue;
+import tech.pegasys.artemis.util.collections.Bitlist;
+
+import java.util.List;
+
+/** {@link Bitlist} accessor */
+public class BitlistAccessor extends AbstractListAccessor {
+
+  @Override
+  public int getChildrenCount(Object value) {
```
mkalinin, line 17: What if we extend general interface by two methods: `getSerialazableUnitsCount` and `getSerializableUnit` with default implementations calling `getChildrenCount` and `getChildValue` respectively?

ssz/src/main/java/org/ethereum/beacon/ssz/type/SSZBitListType.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private long fromAtomicSize(long atomicSize) {
+    int addon = getType() == Type.LIST ? 8 : 7;
+    return atomicSize == SSZType.VARIABLE_SIZE ? atomicSize : (atomicSize + addon) / 8;
+  }
+
+  @Override
+  public long getMaxSize() {
+    return fromAtomicSize(super.getMaxSize());
```
mkalinin, line 54: This call makes a side effect. When lengths is checked against max size it counts bytes required to pack bits of decoding bitlist instead of number of actual bits encoded. It could potentially lead to a false negative result of this check, but taking in account that max size is a power of two we have a very narrow surface on which this side effect takes place, i.e. it could fail only if max size is less than eight bits.

I would not insist but it's highly recommend to keep invariant on `vectorLength` and `maxSize` behaviour. Who knows what other misleading behavior it could lead to.

mkalinin (approved):
Looks good in general. Anyway, I'd like to discuss whether it's reasonable to have fixes for those two places that I've highlighted.

ssz/src/main/java/org/ethereum/beacon/ssz/access/list/BitlistAccessor.java
zilm13: In this way `getChildrenCount` will not be called by anyone. With the same result we could rename this one

ssz/src/main/java/org/ethereum/beacon/ssz/type/SSZBitListType.java
zilm13: It's not a `maxSize` method of Bitlist, it's `maxSize` of ssz type. And the same as above: raw units from it are not needed by anyone on ssz level

ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZSimpleHasher.java
```diff
@@ -72,10 +75,13 @@ public MerkleTrie visitComposite(SSZCompositeType type, Object rawValue,
     } else if ((type.getType() == LIST || type.getType() == VECTOR)
         && ((SSZListType) type).getElementType().getType() == BASIC) {
       SerializerResult sszSerializerResult = serializer.visitAny(type, rawValue);
-      SSZListAccessor listAccessor =
-          (SSZListAccessor) type.getAccessor().getInstanceAccessor(type.getTypeDescriptor());
-      BytesValue serialization =
-          listAccessor.removeListSize(rawValue, sszSerializerResult.getSerializedBody());
+      BytesValue serialization;
+      // Strip size bit in Bitlist
+      if (type.getType() == LIST && type instanceof SSZBitListType) {
```
Nashatyrev: Why didn't you want to add a dedicated types `BITLIST` & `BITVECTOR` ?

Nashatyrev (approved):
This version is much better 👍 
Let it be!

ssz/src/main/java/org/ethereum/beacon/ssz/visitor/SSZSimpleHasher.java
zilm13: This will force to modify all other `== LIST` and `== VECTOR` comparisons to include `BITLIST` and `BITVECTOR` values as Enums doesn't have inheritance or smth like this. I've tried the easy way, check my next commit: https://github.com/harmony-dev/beacon-chain-java/pull/158/commits/e515237ac32e2fc37c9e210befa15458fad25c65

PR #162, opened 2019-08-07
Persistence

mkalinin:
### What was done?
- A new database kind added; it featured with write buffer and designed to work with various storage engines like RocksDB
- Two flushing strategies is implemented:
  - instant: emits flushing on each commit
  - buffer size observer: emits flushing each time when buffer size exceeds predefined limit
- RocksDB source is implemented
- Basic tests for database, rocksdb source and their superposition
- database of `node` executable is switched to RocksDB


Nashatyrev (approved):
Looks nice

zilm13 (changes requested):
LGTM
but tests are failed

start/common/src/main/java/org/ethereum/beacon/start/common/NodeLauncher.java
```diff
@@ -123,7 +124,7 @@ void chainStarted(ChainStart chainStartEvent) {
         new ExtendedSlotTransition(perEpochTransition, perSlotTransition, spec);
     emptySlotTransition = new EmptySlotTransition(extendedSlotTransition);
 
-    db = new InMemoryDatabase();
+    db = Database.rocksDB(Paths.get(computeDbName(chainStartEvent)).toString(), 64L << 20);
```
zilm13: what does this magic number means?

zilm13 (approved):
LGTM

start/common/src/main/java/org/ethereum/beacon/start/common/NodeLauncher.java
mkalinin: Fixed!

PR #163, opened 2019-08-08
Make ordered iteration of tasks scheduled to the same time

Nashatyrev:
Fix non-deterministic behavior of `ControlledScheduler`

### DIFF
--- a/util/src/main/java/org/ethereum/beacon/schedulers/TimeControllerImpl.java
+++ b/util/src/main/java/org/ethereum/beacon/schedulers/TimeControllerImpl.java
@@ -1,16 +1,29 @@
 package org.ethereum.beacon.schedulers;
 
-import com.google.common.collect.SortedMultiset;
 import com.google.common.collect.TreeMultimap;
 import java.util.Comparator;
-import java.util.Iterator;
-import java.util.SortedSet;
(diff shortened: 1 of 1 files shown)


ericsson49 (changes requested):
The new version is a bit less safe.
In the previous version, same task cannot be added twice, since TreeMultimap will ignore it.
However, in the new version, a new instance of OrderedTask will be added, having distinct getOrder() result and thus, it will be different from the TreeMultimap point of view.
It doesn't look like terribly wrong behavior, but it's different from the previous version.

If it's okay, I would also note, that such a doubly added task, won't be removed by a single cancelTask call. This is kinda consistent (added twice then it should be removed twice), but looks unsafe. Previously, some code can call addTask(t) twice for some reason, and a single cancelTask will remove it. But in the new version, there will be a problem.

Nashatyrev:
@ericsson49 adding the same task twice is pretty strange intention. Added javadoc to prevent this :)

Nashatyrev:
The most proper way would be to override `TreeMultimap.createCollection()` to return `LinkedHashSet` instance but it's package private :(

zilm13 (approved):
LGTM

ericsson49 (approved):

PR #164, opened 2019-08-09
Test/db module tests

eustimenko:


db/core/src/test/java/org/ethereum/beacon/db/XorKeyDatabaseTest.java
```diff
@@ -0,0 +1,63 @@
...
+        };
+
+        final DataSource<BytesValue, BytesValue> actual = database.createStorage(TEST_STORAGE_NAME);
+        assertThat(actual).isNotNull();
+        assertThat(actual).isInstanceOf(XorDataSource.class);
+
+        final BytesValue key = BytesValue.wrap(TEST_ENTITY_KEY.getBytes());
+        final BytesValue value = BytesValue.EMPTY;
+        actual.put(key, value);
+        assertThat(actual.get(key)).isPresent().hasValue(value);
```
zilm13: You could create 2 databases with different names and put the same key in both, so you are able to verify that they are not clashed with each other

zilm13 (approved):

db/core/src/test/java/org/ethereum/beacon/db/configuration/EmptyBytesValueGenerator.java
```diff
@@ -0,0 +1,16 @@
+package org.ethereum.beacon.db.configuration;
+
+import com.pholser.junit.quickcheck.generator.*;
```
zilm13: we don't use wildcard imports

db/core/src/test/java/org/ethereum/beacon/db/source/impl/XorDataSourceTest.java
```diff
@@ -0,0 +1,16 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class XorDataSourceTest {
+
+    @Test
+    public void testValidSourceCreation() {
+        assertThatThrownBy(() -> new XorDataSource<>(null, BytesValue.EMPTY))
+                .isInstanceOf(NullPointerException.class);
+    }
+
```
ericsson49, line 16: XorDataSource<>(someDatasource, null) case is missing
No tests for XorDataSource functionality (private static BytesValue xorLongest(BytesValue v1, BytesValue v2) method)

Issue #169, opened 2019-08-14
Eth1 client signing via web3j


zilm13:
Improve Web3j implementation (`pow:web3j`) to sign transactions using client wallet and unlock. Currently we use the same non-secure signer than in other implementations, but with web3j we could move signing to 3rd party client which we call using RPC.

PR #170, opened 2019-08-14
Deposit contract handling using Web3J via JSON-RPC

zilm13:
`pow:web3j` implementation plus some fixes to `pow:ethereumj`


pow/core/src/main/java/org/ethereum/beacon/pow/ContractSource.java
```diff
@@ -0,0 +1,20 @@
+package org.ethereum.beacon.pow;
+
+import java.util.Scanner;
+
+public class ContractSource {
```
mkalinin, line 5: A thought of making this class a singletone. This is rather a cosmetic change than something functional though.

versions.gradle
```diff
@@ -10,7 +10,9 @@ dependencyManagement {
 
     dependency "org.apache.logging.log4j:log4j-api:${log4j2Version}"
     dependency "org.apache.logging.log4j:log4j-core:${log4j2Version}"
+    dependency "org.slf4j:slf4j-simple:1.7.28"
```
mkalinin, line 13: Why do we need `slf4j` if we already have `log4j2`?

pow/web3j/src/test/resources/geth/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Build Geth in a stock Go builder container
+FROM golang:1.12-alpine as builder
+
+RUN apk add --no-cache make gcc musl-dev linux-headers git
+
+ARG branch=master
```
mkalinin, line 6: I think it would be better to stick with some tagged version in order to prevent unnecessary code and/or dependency updates on the `geth` side that could break docker container or the test itself.

pow/web3j/src/main/java/org/ethereum/beacon/pow/contract/DepositContract.java
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * <p>Auto generated code.
+ * <p><strong>Do not modify!</strong>
+ * <p>Please use the <a href="https://docs.web3j.io/command_line.html">web3j command line tools</a>,
+ * or the org.web3j.codegen.SolidityFunctionWrapperGenerator in the 
+ * <a href="https://github.com/web3j/web3j/tree/master/codegen">codegen module</a> to update.
+ *
+ * <p>Generated with web3j version 4.3.2-SNAPSHOT.
+ *
+ * <p>Generated with following input: org.web3j.codegen.SolidityFunctionWrapperGenerator.java -b /home/work/projects/beacon-chain-java/pow/core/src/main/resources/org/ethereum/beacon/pow/ContractBin.bin -a /home/work/projects/beacon-chain-java/pow/core/src/main/resources/org/ethereum/beacon/pow/ContractAbi.json -o /home/work/projects/beacon-chain-java/pow/web3j/src/main/java -p org.ethereum.beacon.pow.contract </p>
```
mkalinin, line 38: We might want to reuse this script later on. Would you mind to include it into a codebase and leave a reference to it here?

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JDepositContract.java
```diff
@@ -0,0 +1,312 @@
...
+    blockExecutor = new LatestExecutor<>(this.schedulers.blocking(), this::processBlocksUpTo);
+    this.web3RequestExecutor = new Web3RequestExecutor(web3j);
+  }
+
+  @Override
+  protected void chainStartSubscribed() {
+    web3RequestExecutor.executeOnSyncDone(
+        () -> {
+          processConfirmedBlocks();
+          onEthereumUpdated();
```
mkalinin, line 87: `onEthereumUpdated()` in its turn calls to `processConfirmedBlocks()`. Is that by design?

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JDepositContract.java
```diff
@@ -0,0 +1,312 @@
...
+        });
+    web3j
+        .blockFlowable(false)
+        .doOnNext(
+            block -> {
+              onEthereumUpdated();
+            })
+        .subscribe();
```
mkalinin, line 95: Does that mean that we're subscribed to blocks from very beginning before sync is done? We might want to get this subscription right on sync done event. Is that doable with web3j? Probable obstacle that I can see so far is that sync done event is not always emitted. For example, would it be emitted if sync was done before we get subscribed?

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JDepositContract.java
```diff
@@ -0,0 +1,312 @@
...
+    blockExecutor.newEvent(bestConfirmedBlock);
+  }
+
+  void processBlocksUpTo(long bestConfirmedBlock) {
+    for (long number = processedUpToBlock; number <= bestConfirmedBlock; number++) {
+      EthBlock block =
+          web3j
+              .ethGetBlockByNumber(DefaultBlockParameter.valueOf(BigInteger.valueOf(number)), false)
+              .sendAsync()
+              .join();
```
mkalinin, line 123: I am thinking of processing blocks and transaction receipts in an asynchronous fashion. We probably can get a block flow out of web3j, then add a processor that filters out blocks with usage of bloom filter, then yet another one which requests for tx receipts of valuable blocks. I guess we can created a stream upon a set of futures that are requesting for receipts. And at the end parse deposit data from tx receipts upon their arrival. 

This scheme has at least one flaw, deposit events might be processed in the wrong order but I don't think that there will be a big mess. So we could probably have a buffer that could be drained when there is a deposit standing next to last processed one.

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JTransactionBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+                  .ethGetTransactionCount(fromAddress, DefaultBlockParameterName.LATEST)
+                  .sendAsync();
+          CompletableFuture<EthGasPrice> gasPriceFut = web3j.ethGasPrice().sendAsync();
+          CompletableFuture.allOf(txCountFut, gasPriceFut)
+              .thenAccept(
+                  aVoid -> {
+                    BigInteger nonce = txCountFut.join().getTransactionCount();
+                    BigInteger value =
+                        Convert.toWei(amount.longValue() + "", Convert.Unit.GWEI).toBigInteger();
+                    BigInteger gasLimit = BigInteger.valueOf(2_000_000);
```
mkalinin, line 62: Let's create a constant for this magic number.

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JTransactionBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+                    result.complete(BytesValue.wrap(rawTransaction));
+                  });
+        });
+
+    return result;
+  }
+
+  @Override
+  public CompletableFuture<BytesValue> signTransaction(
+      BytesValue unsignedTransaction, BytesValue eth1PrivKey) {
```
mkalinin, line 93: We might want to avoid private key management and dedicate it to geth or parity. Any thoughts on that?

pow/web3j/src/test/resources/geth/README.md
```diff
@@ -0,0 +1,14 @@
+Go-Ethereum instance with open RPC and non-persistent DB initialized with `genesis.json`.   
+
+Building image:
+```shell script
+docker build .
+```
+Running container:
+```shell script
+docker run -d -p 8545:8545 -p 30303:30303 <image id>
```
mkalinin, line 9: It would be better to give this image a name: `docker build -t name && docker run -d name`.

mkalinin (changes requested):
Great job is done! I left a few requests for improvement.

pow/web3j/src/main/java/org/ethereum/beacon/pow/Web3JTransactionBuilder.java
zilm13: I've created issue #169

pow/web3j/src/test/java/org/ethereum/beacon/pow/AbstractWeb3JTest.java
```diff
@@ -0,0 +1,124 @@
...
+                return Time.of(0);
+              }
+            })
+        .withDefaultHasher()
+        .build();
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    this.web3j = Admin.build(new HttpService());
```
mkalinin, line 103: Since RPC URL may vary from one env to another it would be better if it's explicitly passed to web3j.

PR #175, opened 2019-08-26
Encapsulate batch writer logic into WriteBuffer

mkalinin:
### What was done?
- Move batch writer logic to `WriteBuffer`
- Remove standalone `BatchWriter` class


ericsson49 (approved):
LGTM

db/core/src/main/java/org/ethereum/beacon/db/source/WriteBuffer.java
```diff
@@ -88,12 +88,12 @@ public void remove(@Nonnull final K key) {
...
-      for (Map.Entry<K, CacheEntry<V>> entry : buffer.entrySet()) {
-        if (entry.getValue() == CacheEntry.REMOVED) {
-          getUpstream().remove(entry.getKey());
-        } else {
-          getUpstream().put(entry.getKey(), entry.getValue().value);
-        }
+      if (getUpstream() instanceof BatchUpdateDataSource) {
+        final Map<K, V> updates = new HashMap<>();
+        buffer.forEach((key, value) -> updates.put(key, value.value));
+        ((BatchUpdateDataSource<K, V>) getUpstream()).batchUpdate(updates);
```
Nashatyrev, line 94: What about `REMOVED` entries?

mkalinin: That's a good question. `REMOVED` cache entry has a value equal to `null` which preserves a null-removed contract. I'll add a comment to `batchUpdate` and `REOMVED` entry about that.

Nashatyrev (approved):

db/core/src/main/java/org/ethereum/beacon/db/source/WriteBuffer.java
mkalinin: Fixed https://github.com/harmony-dev/beacon-chain-java/pull/175/commits/c65f5c05995cd44871ffd322de67fb0471953586.

eustimenko (approved):

PR #176, opened 2019-08-28
[WIP] Attestation pool take 1: interop

mkalinin:


Nashatyrev:
`TimeProcessors` and similar classes: 
what's the rationale behind merging all events into single stream and then sorting them out by `instanceof`? 
May be it make sense to subscribe all `feed*` methods directly to the corresponding streams? (to make this thread safe you may do `publishOn(singleThreadScheduler)` for all inbound streams)

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/IdentificationProcessor.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Override
+  protected void hookOnNext(Object value) {
+    if (value.getClass().equals(BeaconBlock.class)) {
+      // forward attestations identified with a new block
+      pool.feedNewImportedBlock((BeaconBlock) value).forEach(this::publishOut);
+    } else if (value.getClass().equals(SlotNumber.class)) {
+      pool.feedNewSlot((SlotNumber) value);
+    } else if (value.getClass().equals(ReceivedAttestation.class)) {
+      if (pool.isInitialized()) {
```
Nashatyrev: shouldn't it be 
```java
if (!pool.isInitialized()) 
```

chain/src/main/java/org/ethereum/beacon/chain/pool/StatefulProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+ * called by its clients.
+ *
+ * <p>{@link #isInitialized()} method indicates whether processor has already been initialized or
+ * not. It's a client responsibility to check {@link #isInitialized()} result before calling to the
+ * instance of this processor.
+ *
+ * <p>Implementor MAY throw an {@link AssertionError} if it's been called before inner state has
+ * been initialised.
+ */
+public interface StatefulProcessor {
```
Nashatyrev, line 16: Not sure I understand the purpose of this interface. 
You either `assert` that a component is inited or just skip an entry if not inited. Couldn't this be the internal logic of this component?

Nashatyrev:
Regarding a component which is statefull and has several inbound streams - I'm still not sure what should be the proper pattern for it
Ideally all this RX stuff should be stateless but often it is pretty tricky and we need to go back to imperative world

chain/src/main/java/org/ethereum/beacon/chain/pool/StatefulProcessor.java
mkalinin: For instance, we have checkers which returns value of `boolean` type saying whether check passed or not. These checkers may have several inputs and state initialized from either one or all of those inputs. Since, check MUST returns `true/false` there is no room for merely discarding the value fed to a checker. We might use `Optional<Boolean>` for that but I am not sure that it would be a better choice.

mkalinin:
> Ideally all this RX stuff should be stateless but often it is pretty tricky and we need to go back to imperative world

I agree with that. It looks like Reactor is not meant to handle stateful processors and was created for another purposes. This way it looks more like a misuse. Let's keep it for now as it is and replace Reactor with a proper thing later.

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/IdentificationProcessor.java
```diff
@@ -24,46 +25,49 @@
...
-  private final DirectProcessor<ReceivedAttestation> unknownAttestations = DirectProcessor.create();
-  private final FluxSink<ReceivedAttestation> unknownOut = unknownAttestations.sink();
+  private final OutsourcePublisher<ReceivedAttestation> identified = new OutsourcePublisher<>();
+  private final OutsourcePublisher<ReceivedAttestation> unknown = new OutsourcePublisher<>();
 
-  public IdentificationProcessor(UnknownAttestationPool pool) {
+  public IdentificationProcessor(
+      UnknownAttestationPool pool,
+      Schedulers schedulers,
+      Flux<ReceivedAttestation> source,
```
Nashatyrev: I would consider using more general `Publisher` as an input. I believe `Flux.from(Publisher)` shortcuts the call when the parameter is `Flux` so this should be for free

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/IdentificationProcessor.java
```diff
@@ -24,46 +25,49 @@
...
       }
-    } else {
-      throw new IllegalArgumentException(
-          "Unsupported input type: " + value.getClass().getSimpleName());
     }
   }
 
-  public DirectProcessor<ReceivedAttestation> getUnknownAttestations() {
-    return unknownAttestations;
+  public OutsourcePublisher<ReceivedAttestation> getIdentified() {
```
Nashatyrev: I wouldn't expose `OutsourcePublisher` implementation until the caller is supposed to write items to this stream. Wouldn't returning `Publisher` suits?

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/SanityProcessor.java
```diff
@@ -1,47 +1,66 @@
...
+    if (checker.check(attestation)) {
+      valid.publishOut(attestation);
     } else {
-      throw new IllegalArgumentException(
-          "Unsupported input type: " + value.getClass().getSimpleName());
+      invalid.publishOut(attestation);
     }
   }
+
+  public Flux<ReceivedAttestation> getValid() {
```
Nashatyrev: The same here, m.b. just `Publisher` ?

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/SignatureEncodingProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public Flux<ReceivedAttestation> getValid() {
+    return valid;
+  }
+
+  public Flux<ReceivedAttestation> getInvalid() {
+    return invalid;
+  }
+}
```
Nashatyrev, line 59: Why didn't you finally end up with the laconic form as we discussed? 
```java
validatedAtt = att.map(att -> Pair.of(att, checker.check(att))).publish()
validAtt = validatedAtt.filter(a -> a.right).map(a -> a.left)
invalidAtt = validatedAtt.filter(a -> !a.right).map(a -> a.left)
```

chain/src/main/java/org/ethereum/beacon/chain/pool/reactor/TimeProcessor.java
```diff
@@ -1,51 +1,59 @@
...
+  private void hookOnNext(ReceivedAttestation attestation) {
+    if (filter.isInitialized() && filter.check(attestation)) {
+      out.publishOut(attestation);
     }
   }
+
+  @Override
+  public void subscribe(CoreSubscriber<? super ReceivedAttestation> actual) {
+    out.subscribe(actual);
+  }
```
Nashatyrev, line 66: Why not expose the `Publisher<ReceivedAttestation>` here to make it uniform across the whole lib?

Nashatyrev:
The general feeling is that using RX shouldn't require such amount of boilerplate code

PR #177, opened 2019-08-29
Feature/mocked start

ericsson49:
1. Mocked start additions implemented.
- eth1 block hash
- key pairs generation
- withdrawal credentials
- yaml spec constants file can be specified as a cmd line argument
- default config updated
2. rocks DB JNI version changed

Starting from a specified state is not implemented.


mkalinin (approved):
Looks good! 👍 I made a tiny aesthetic fix. Would like to merge these changes asap.

PR #182, opened 2019-09-06
Feature/mocked start

ericsson49:
Support for loading beacon state files added (SSZ format).

### DIFF
--- a/start/common/src/main/java/org/ethereum/beacon/start/common/util/StorageUtils.java
+++ b/start/common/src/main/java/org/ethereum/beacon/start/common/util/StorageUtils.java
@@ -1,0 +1,46 @@
+package org.ethereum.beacon.start.common.util;
+
+import org.ethereum.beacon.chain.BeaconTuple;
+import org.ethereum.beacon.chain.storage.BeaconChainStorage;
+import org.ethereum.beacon.consensus.BeaconChainSpec;
+import org.ethereum.beacon.consensus.BeaconStateEx;
+import org.ethereum.beacon.core.BeaconBlock;
+import org.ethereum.beacon.core.state.Checkpoint;
+import tech.pegasys.artemis.ethereum.core.Hash32;
+
+/**
+ * Utility functions to initialize storage from an initial state.
+ */
+public class StorageUtils {
+  /**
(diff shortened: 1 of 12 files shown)


start/node/src/main/java/org/ethereum/beacon/node/Node.java
```diff
@@ -109,6 +109,44 @@
...
+  private String startMode;
+
+  @CommandLine.Option(
+      names = "--force-db-clean",
+      paramLabel = "force-db-clean",
+      description = {
+          "When an initial-state is specified, but db is not empty",
+          "specifies how to resolve the problem:",
+          "  force-db-clean=true - tries to clean db",
+          "  force-db-clean=true - exits with failure status.",
```
mkalinin: There should be `false` here, shouldn't it?

start/common/src/main/java/org/ethereum/beacon/start/common/util/StorageUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import org.ethereum.beacon.consensus.BeaconChainSpec;
+import org.ethereum.beacon.consensus.BeaconStateEx;
+import org.ethereum.beacon.core.BeaconBlock;
+import org.ethereum.beacon.core.state.Checkpoint;
+import tech.pegasys.artemis.ethereum.core.Hash32;
+
+/**
+ * Utility functions to initialize storage from an initial state.
+ */
+public class StorageUtils {
```
mkalinin: We might want to make this util a part of `chain` module to make it self-sufficient in terms of chain initialization.

mkalinin (approved):
Looks good in general. A couple of things to fix and it could be merged.

start/node/src/main/java/org/ethereum/beacon/node/Node.java
ericsson49: Definitely!

start/common/src/main/java/org/ethereum/beacon/start/common/util/StorageUtils.java
ericsson49: Yes. This also will help fixing failed tests.

PR #186, opened 2019-09-12
WIP: Dockerfile for genesis state start

atoulme:


zilm13:
@atoulme thank you for initial commit, I will finish task here: #187

PR #189, opened 2019-09-16
Discovery v5 

zilm13:
Implementation of this:
https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md

Still in work, need to finish some things, add network, plus refactoring especially around messageHandler <-> Context, the system is raw.


zilm13:
Network using Netty added.

TODO:
- make it work w/o network (network looks not an issue)
- polish

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+
+import java.math.BigInteger;
+import java.util.Base64;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Ethereum Node Record
+ *
+ * <p>Node record as described in <a href="https://eips.ethereum.org/EIPS/eip-778>EIP-778</a>
```
mkalinin: Missed closing quote in `href` value.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+    RlpString id = (RlpString) values.get(2);
+    if (!"id".equals(new String(id.getBytes()))) {
+      throw new RuntimeException(
+          String.format(
+              "Unable to deserialize ENR with no id field at 2-3 records, [%s]",
+              BytesValue.wrap(bytes)));
+    }
+    RlpString idVersion = (RlpString) values.get(3);
+
+    Function<List<RlpType>, NodeRecord> nodeRecordV4Creator =
```
mkalinin: What if we delegate v4 and v5 creators to corresponding classes?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+        fields -> {
+          NodeRecordV4.Builder builder =
+              NodeRecordV4.Builder.empty()
+                  .withSignature(BytesValue.fromHexString(((RlpString) values.get(0)).asString()))
```
mkalinin: Converting to string and then parsing that string back looks redundant. What is the rationale behind this redundancy?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Ethereum Node Record
+ *
+ * <p>Node record as described in <a href="https://eips.ethereum.org/EIPS/eip-778>EIP-778</a>
+ */
+@SSZSerializable(listAccessor = NodeRecord.NodeRecordAccessor.class)
```
mkalinin: Discv5 is tightly coupled with RLP. Can't we leverage from that and do not bring SSZ to this module?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+        {
+          return nodeRecordV4Creator.apply(values);
+        }
+      case V5:
+        {
+          return nodeRecordV5Creator.apply(values);
+        }
+    }
+
+    return null;
```
mkalinin: Let's use `@Nullable` annotation for those methods which explicitly returns `null`.

zilm13:
TODO: Buckets according to spec
TODO: handshake crypto

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Ethereum Node Record
+ *
+ * <p>Node record as described in <a href="https://eips.ethereum.org/EIPS/eip-778>EIP-778</a>
+ */
+@SSZSerializable(listAccessor = NodeRecord.NodeRecordAccessor.class)
+public interface NodeRecord {
```
mkalinin: This interface should provide an access to arbitrary ENR pairs that are described in the EIP. Representing list of pairs as class fields doesn't seem to be extensible. IMO, it should be a kind of map.

mkalinin: We can have a list of methods and/or enum values corresponding to pairs defined by EIP.

mkalinin (changes requested):

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: > This interface should provide an access to arbitrary ENR pairs that are described in the EIP

As I understand from EIP only `signature` and `seq` are arbitrary. I've added both.

zilm13: > We can have a list of methods and/or enum values corresponding to pairs defined by EIP.

I will add it soon

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: Made it throw `Exception` instead

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: fixed

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: fixed

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: fixed

zilm13:
- TODO: task queue for session (only one task at time currently)
- TODO: FIXME & TODOs
- TODO: all handshake auth correctly

zilm13:
@mkalinin ready for review
Current TODO:
- more tests (bad packets/messages handling, 2+ tasks enqueued etc)
- go discv5 implementation compliance test

discovery/src/main/java/org/ethereum/beacon/discovery/enr/EnrSchemeInterpreter.java
```diff
@@ -0,0 +1,23 @@
+package org.ethereum.beacon.discovery.enr;
+
+import org.web3j.rlp.RlpString;
+import tech.pegasys.artemis.util.bytes.Bytes32;
+
+public interface EnrSchemeInterpreter {
```
mkalinin: I think `IdentitySchema` name should be closer to the domain.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/EnrSchemeInterpreter.java
```diff
@@ -0,0 +1,23 @@
...
+  default void verify(NodeRecord nodeRecord) {
+    if (!nodeRecord.getIdentityScheme().equals(getScheme())) {
+      throw new RuntimeException("Interpreter and node record schemes do not match!");
+    }
+  }
+
+  /** Delivers nodeId according to identity scheme scheme */
+  Bytes32 getNodeId(NodeRecord nodeRecord);
+
+  Object decode(String key, RlpString rlpString);
```
mkalinin: Identity scheme is not responsible for encode/decode. Why would we want to add this responsibility here?

If you want to make an abstraction for ENR codec you may do this with a separated interface. The only implementation that makes sense so far is an RLP codec.

Discovery might be initialized with identity scheme factory that in its turn is initialized with a list of supported schemes. When `NodeRecord` gets parsed it's passed on identity schema to get verified; also, identity scheme is used to derive node id. It could be done in a factory method as you already do it in the following code.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/EnrSchemeV4Interpreter.java
```diff
@@ -0,0 +1,123 @@
...
+      return RlpString.create(new byte[0]);
+    } else {
+      throw new RuntimeException(
+          String.format(
+              "Couldn't serialize node record field %s with value %s: no serializer found.",
+              key, object));
+    }
+  }
+
+  private RlpString fromNumber(Number number) {
```
mkalinin: I think that it's better to place these methods to `RlpUtils`.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Set;
+
+/**
+ * Ethereum Node Record
+ *
+ * <p>Node record as described in <a href="https://eips.ethereum.org/EIPS/eip-778">EIP-778</a>
+ */
+public class NodeRecord {
+  // Compressed secp256k1 public key, 33 bytes
+  public static String FIELD_PKEY_SECP256K1 = "secp256k1";
```
mkalinin: As a cleanup we may put these constants to a separate interface.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,192 @@
...
+  // UDP port, integer
+  public static String FIELD_UDP_V4 = "udp";
+  // IPv6 address
+  public static String FIELD_IP_V6 = "ip6";
+  // IPv6-specific TCP port
+  public static String FIELD_TCP_V6 = "tcp6";
+  // IPv6-specific UDP port
+  public static String FIELD_UDP_V6 = "udp6";
+
+  private UInt64 seq;
```
mkalinin: Why not `final`?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,192 @@
...
+  // Signature
+  private BytesValue signature;
+  // optional fields
+  private Map<String, Object> fields = new HashMap<>();
+
+  private EnrSchemeInterpreter enrSchemeInterpreter;
```
mkalinin: Creating a scheme doesn't have big resource impact; moreover, it happens infrequent. Let's maybe do this on demand and get rid of this field?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,192 @@
...
+
+  public void verify() {
+    enrSchemeInterpreter.verify(this);
+  }
+
+  public RlpList asRlp() {
+    return asRlp(true);
+  }
+
+  public RlpList asRlp(boolean withSignature) {
```
mkalinin: I think to increase readability it's better to create two shortcuts `asRlp` and `asRlpNoSignature`. Then reader won't need to check what's this boolean parameter for.

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
```diff
@@ -0,0 +1,192 @@
...
+    return new RlpList(values);
+  }
+
+  public BytesValue serialize() {
+    return serialize(true);
+  }
+
+  public BytesValue serialize(boolean withSignature) {
+    byte[] bytes = RlpEncoder.encode(asRlp(withSignature));
+    assert bytes.length <= 300;
```
mkalinin: Magic number?

discovery/src/main/java/org/ethereum/beacon/discovery/enr/NodeRecord.java
zilm13: I already use this for tests to have two interpreters: with verification and without. I see no better approach while having feature of verification override.

zilm13:
@mkalinin Finished it. Please, review.

shahankhatch:
lgtm. I have a minor request to review/accept #210 which renames two methods.

discovery/src/test/java/org/ethereum/beacon/discovery/DiscoveryInteropTest.java
```diff
@@ -0,0 +1,121 @@
...
+@Ignore("Requires manual startup, takes a bit to start")
+public class DiscoveryInteropTest {
+  @Test
+  public void testInterop() throws Exception {
+    // 1) start 2 nodes
+    Pair<BytesValue, NodeRecord> nodePair1 = TestUtil.generateNode(40412, true);
+    System.out.println(String.format("Node %s started", nodePair1.getValue1().getNodeId()));
+    NodeRecord nodeRecord1 = nodePair1.getValue1();
+    NodeRecord nodeRecord2 =
+        NODE_RECORD_FACTORY_NO_VERIFICATION.fromBase64(
```
mkalinin: It would be convenient if one may change parameters of this node like IP address. I had to replace this base64 line to make test run on Mac OS as it handles docker in a `docker-machine` instance with its own network interface and addresses.

mkalinin (commented):
I am in favor of merging this PR and making further refactor/updates via separate requests. I have a suggestion to rename `discovery` module to `discv5` for clarity.
@shahankhatch @zilm13 WDYT?

P.S. interop test looks fantastic! There is a little fix that could increase its usability.

mkalinin (approved):

zilm13:
@mkalinin I don't like `discv5` as a lot of code could be reused in, say, `discv6`

shahankhatch:
> @mkalinin I don't like `discv5` as a lot of code could be reused in, say, `discv6`

I agree with this reasoning.

mkalinin:
This particular implementation is `discv5`. I don't think it's gonna be re-used for `discvX` in the future. Anyway, it doesn't matter much :)

discovery/src/test/java/org/ethereum/beacon/discovery/DiscoveryInteropTest.java
zilm13: Yeah, did this, plus refactored a bit NodeRecord creation

Issue #196, opened 2019-09-18
ForkChoice::get_head doesn't break ties properly


ericsson49:
[fork_choice/get_head](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#get_head) specifies, that when choosing the heaviest child, ties should be broken lexicographically:
> \# Sort by latest attesting balance with ties broken lexicographically

However, Harmony implementation sorts by latest balance only.
`head = children.stream().max(Comparator.comparing(root -> get_latest_attesting_balance(store, root))).get();`

Issue #191, opened 2019-09-17
Fork choice specs and implementation issues: blocks


ericsson49:
List of differences, that I found, between [Fork choice specs](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md) and our implementation. This doesn't necessarily mean it's a bug.

1 Blocks are not delayed and time check is different
Spec says:
> \# Blocks cannot be in the future. If they are, their consideration must be delayed until the are in the past.
>  `assert store.time >= pre_state.genesis_time + block.slot * SECONDS_PER_SLOT`

However, DefaultBeaconChain::insert simply rejects blocks too far the future, using a bit different time check:
`assert block.slot <= get_current_slot(store.time) + 1`

2 Blocks are not checked against finalized checkpoint
Spec says:
> \# Check block is a descendant of the finalized block

And also
> \# Check that block is later than the finalized epoch slot

The time check seems to be redundant, given the block passes the descendance check.
Our implementation performs a similar check against justified checkpoint `LMDGhostProcessor::isJustifiedAncestor`.
The other issue is that it's performed after the block is imported.
While, according to the spec, it's not imported if the finality-descendance check is failed.

3 Finality updates

Spec says:
```
# Update justified checkpoint
if state.current_justified_checkpoint.epoch > store.justified_checkpoint.epoch:
    store.justified_checkpoint = state.current_justified_checkpoint
# Update finalized checkpoint
if state.finalized_checkpoint.epoch > store.finalized_checkpoint.epoch:
    store.finalized_checkpoint = state.finalized_checkpoint
```

However, DefaultBeaconChain::updateFinality implementation is somewhat different:

```
    if (!previous.getFinalizedCheckpoint().equals(current.getFinalizedCheckpoint())) {
      chainStorage.getFinalizedStorage().set(current.getFinalizedCheckpoint());
      finalizedCheckpointStream.onNext(current.getFinalizedCheckpoint());
    }
    if (!previous.getCurrentJustifiedCheckpoint().equals(current.getCurrentJustifiedCheckpoint())) {
      // store new justified checkpoint if its epoch greater than previous one
      if (current
          .getCurrentJustifiedCheckpoint()
          .getEpoch()
          .greater(fetchJustifiedCheckpoint().getEpoch())) {
        chainStorage.getJustifiedStorage().set(current.getCurrentJustifiedCheckpoint());
      }
```

[...]

PR #199, opened 2019-09-18
Fix/fork choice issues

ericsson49:
1. #196 fixed
2. partial fix of #191 - update finality fixed


consensus/src/main/java/org/ethereum/beacon/consensus/spec/ForkChoice.java
```diff
@@ -109,7 +110,7 @@ default Hash32 get_head(Store store) {
       }
 
       head = children.stream()
-          .max(Comparator.comparing(root -> get_latest_attesting_balance(store, root)))
+          .max(Comparator.comparing(root -> Pair.with(get_latest_attesting_balance(store, root), root)))
```
mkalinin, line 113: Let's make sure that implementation of `BytesValue.compareTo` is aligned with the spec.

ericsson49: Compared to python.
Python spec implementation uses bytes (python type) to hold hash32 values.
They are compared with a lexicographical order which matches BytesValue implementation.
E.g. from left to right, unsigned bytes.
Added a test for that.

PR #209, opened 2019-11-04
Helper methods to create updated attestation.

ericsson49:


core/src/main/java/org/ethereum/beacon/core/operations/Attestation.java
```diff
@@ -79,6 +81,24 @@ public BLSSignature getSignature() {
     return signature;
   }
 
+  public Attestation withAggregationBits(Bitlist aggregationBits, SpecConstants specConstants) {
+    return new Attestation(
+        ensureMaxSize(aggregationBits, specConstants), data, custodyBits, BLSSignature.ZERO);
+  }
+
+  public Attestation withData(AttestationData data) {
+    return new Attestation(aggregationBits, data, custodyBits, BLSSignature.ZERO);
```
mkalinin, line 92: What if we add an explicit check like `assert signature == BLSSignagure.ZERO`?

ericsson49: To forbid updating already signed attestations?

mkalinin: Signed attestation looks like an invariant that we might want to preserve. WDYT?

ericsson49: It makes sense. Updated the code.

mkalinin (approved):
LGTM! 👍

PR #210, opened 2019-11-23
Refactor method name to distinguish by parameter name

shahankhatch:
This refactor is to make it clearer when magic is applied to the first parameter in `WhoAreYouPacket`. For both `create` methods, I thought the `Bytes` and `Bytes32` types of the first parameter were too closely related, making it hard for me to distinguish when generalizing types (integration of this discv5 into Artemis uses Tuweni `Bytes`).


zilm13:
Cool, I  like it

